PR #49, opened 2026-01-21
Fix Stop hook path resolution when CLAUDE_PLUGIN_ROOT is not set

fahmyelraie:
The Stop hook was failing with "No such file or directory" error when CLAUDE_PLUGIN_ROOT was not properly set in the environment, resulting in the script path resolving to "/scripts/check-complete.sh" instead of the full path.

Changes:
- Add SCRIPT_DIR variable with fallback to explicit path
- Use $HOME/.claude/plugins/planning-with-files/scripts as fallback when CLAUDE_PLUGIN_ROOT is not set
- Make check-complete.sh executable (chmod +x)
- Apply fix to both Unix and Windows paths in the hook command

This ensures the Stop hook runs successfully even when CLAUDE_PLUGIN_ROOT is not available in the shell environment.

Fixes the error: "bash: /scripts/check-complete.sh: No such file or directory"


OthmanAdi:
Thank you @fahmyelraie for this excellent fix! 🙏

This resolves a real issue that users were experiencing when `CLAUDE_PLUGIN_ROOT` wasn't properly set. The fallback path approach is clean and defensive.

Merged and will be included in the next release (v2.6.0)!

PR #51, opened 2026-01-22
Add the start command

Guozihong:
- Add the start command, which can be used through /planning-with-files:start in claudecode without the need to copy skills
- Copying skills to .claude/skills can also lead to ${CLAUDE-PLUGN_ROOT} path errors


OthmanAdi:
Thank you @Guozihong for this contribution! 🙏

This makes installation much simpler - users can now use `/planning-with-files:start` directly without needing to copy skills to their local folder.

Merged and will be included in the next release (v2.6.0)!

PR #76, opened 2026-02-09
Add opencode scripts for planning-with-files skill

codelyc:
Change-Id: I0f63b88df432ebae3a32e9037e7217623adadf11


OthmanAdi:
Thanks for taking the time to add OpenCode support. The scripts look solid and fill a real gap for users on that platform. Appreciate the contribution.

PR #86, opened 2026-02-27
fix: Stop hook fails on Windows — multiline YAML command not parsed by Git Bash

raykuo998:
## Summary

- The Stop hook in SKILL.md uses a YAML multiline block (`command: |`) with a ~25-line bash script for OS detection. On Windows, Claude Code passes hook commands through Git Bash, which fails to parse the multiline content — `SCRIPT_DIR=...` is interpreted as a command name instead of a variable assignment
- Replaced the multiline block with a single-line command: set `SD` variable inline, try `powershell.exe` first (Windows), fall back to `sh` (Linux/macOS). The OS-detection logic is eliminated entirely since the fallback chain handles it implicitly
- Applied to all 7 tool variants (claude, cursor, codex, codebuddy, kilocode, mastracode, opencode)

## Error on Windows (before fix)

```
Stop hook error: Failed with non-blocking status code: bash:
SCRIPT_DIR=/c/Users/rayku/.claude/skills/planning-with-files/scripts: No such file or directory
sh: /check-complete.sh: No such file or directory
```

## Root Cause

YAML `command: |` multiline blocks are not reliably parsed by Git Bash on Windows. The shell receives the first line (`SCRIPT_DIR=...`) as a command to execute rather than a variable assignment, causing the entire hook to fail.

## Fix

Before (25 lines per variant):
```yaml
command: |
  SCRIPT_DIR="${CLAUDE_PLUGIN_ROOT:-$HOME/.claude/plugins/planning-with-files}/scripts"
  IS_WINDOWS=0
  if [ "${OS-}" = "Windows_NT" ]; then
    ...20 more lines of OS detection...
  fi
```

After (1 line per variant):
```yaml
command: "SD=\"${CLAUDE_PLUGIN_ROOT:-$HOME/.claude/plugins/planning-with-files}/scripts\"; powershell.exe -NoProfile -ExecutionPolicy Bypass -File \"$SD/check-complete.ps1\" 2>/dev/null || sh \"$SD/check-complete.sh\""
```

## Test plan

- [x] Verified fix resolves the error on Windows 11 with Git Bash
- [x] Verify `check-complete.ps1` runs correctly via powershell.exe on Windows
- [x] Verify `check-complete.sh` runs correctly via sh on Linux/macOS (powershell.exe not found, falls through to sh)

### DIFF
--- a/.mastracode/skills/planning-with-files/SKILL.md
+++ b/.mastracode/skills/planning-with-files/SKILL.md
@@ -17,7 +17,7 @@ hooks:
   Stop:
     - hooks:
         - type: command
-          command: "SD=\"$HOME/.mastracode/skills/planning-with-files/scripts\"; powershell.exe -NoProfile -ExecutionPolicy Bypass -File \"$SD/check-complete.ps1\" 2>/dev/null || sh \"$SD/check-complete.sh\""
+          command: "SD=\"$HOME/.mastracode/skills/planning-with-files/scripts\"; [ -f \"$SD/check-complete.sh\" ] || SD=\".mastracode/skills/planning-with-files/scripts\"; powershell.exe -NoProfile -ExecutionPolicy Bypass -File \"$SD/check-complete.ps1\" 2>/dev/null || sh \"$SD/check-complete.sh\""
 metadata:
   version: "2.18.1"
 ---


OthmanAdi:
Ill do the last check as soon as possible. Thank you so much for the contributions 🙏🙏 @raykuo998

OthmanAdi:
Merged. Thank you @raykuo998.

Two PRs on the same day, both targeting Windows, both tracking bugs all the way to root cause and fixing them clean across every platform copy.

PR #88 fixed check-complete.ps1 completely failing on PowerShell 5.1. The parser was choking on characters inside double-quoted Write-Host strings: square brackets treated as array index expressions, parentheses as subexpressions, and the em-dash in comments breaking the block entirely. Switching to single-quoted strings with explicit concatenation was the correct call, and you applied it across all 12 copies without missing one.

This PR fixed the Stop hook YAML multiline block not parsing under Git Bash. The shell was receiving the first line as a command name instead of a variable assignment, crashing the hook before it could do anything. Collapsing 25 lines of OS detection into a single-line implicit fallback chain is cleaner and more maintainable than what was there, and it works correctly on every platform.

Windows users of this skill now have a working Stop hook and a working completion script. That matters more than most people realize.

Tagging v2.18.3 with both fixes credited. Welcome to the contributors list.

Ahmad

PR #100, opened 2026-03-04
Codex catchup: avoid Claude-path silent failure

### LINKED ISSUE #94: Codex variant: session-catchup still reads Claude session path
I was testing the Codex variant and noticed the catchup script here:

- `.codex/skills/planning-with-files/scripts/session-catchup.py`

still resolves sessions from `~/.claude/projects` (Claude layout).

For Codex users, this means catchup can look in the wrong place and miss previous session context after `/clear`.

What I saw
- `get_project_dir()` in the Codex script is hardcoded to Claude storage.

Suggested fix
- Detect provider/storage path in the Codex variant, or
- If Codex parsing isn't implemented yet, print an explicit fallback message instead of scanning Claude paths.

Happy to open a PR once you confirm the intended Codex session path.

tt-a1i:
## Summary
Fixes Codex session-catchup behavior so it no longer silently scans Claude's session path.

### What changed
- Updated `.codex/skills/planning-with-files/scripts/session-catchup.py`.
- When running from Codex skill folder and `~/.codex/sessions` is present (while the Claude project path is absent), the script now prints an explicit fallback message and exits.
- This avoids the previous silent no-op caused by looking at `~/.claude/projects/...` for Codex sessions.

## Why
Codex stores session data under `~/.codex/sessions` in a different format. The previous behavior used Claude path assumptions and could miss context recovery without any explanation.

## Validation
- Ran script with no planning files: no output (as expected).
- Ran script with planning files present: fallback message printed clearly.

Closes #94

### DIFF
--- a/.codex/skills/planning-with-files/scripts/session-catchup.py
+++ b/.codex/skills/planning-with-files/scripts/session-catchup.py
@@ -13,18 +13,31 @@ import sys
 import os
 from pathlib import Path
 from typing import List, Dict, Optional, Tuple
-from datetime import datetime
 
 PLANNING_FILES = ['task_plan.md', 'progress.md', 'findings.md']
 
 
-def get_project_dir(project_path: str) -> Path:
-    """Convert project path to Claude's storage path format."""
+def get_project_dir(project_path: str) -> Tuple[Optional[Path], Optional[str]]:
+    """Resolve session storage path for the current runtime variant."""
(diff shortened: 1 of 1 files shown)


OthmanAdi:
@tt-a1i The fix is correct. Detecting the Codex variant from `__file__` is clean — no config needed, no ambiguity. The explicit fallback message is much better than a silent no-op. Merging now.

PR #109, opened 2026-03-19
fix: restore Codex session-catchup fallback

YSAA1:
## Summary

  This restores the Codex-specific `session-catchup.py` fallback that was previously added in PR #100, but appears to have been overwritten by a later sync/update.

  ## Problem

  PR #100 added a Codex safeguard so the Codex variant would not silently scan Claude session storage (`~/.claude/projects`) when running from the Codex skill path.

  However, in the current codebase, the Codex-distributed `session-catchup.py` no longer contains that fallback logic. As a result, the Codex variant can again fall back to Claude-path
  behavior instead of explicitly skipping with a clear message.

  ## Root Cause

  From tracing the history locally:

  - PR #100 merged the Codex-specific fallback into the distributed Codex script
  - Later updates (including the v2.23.0 session-catchup changes) replaced the skill-distributed script contents from the shared source
  - The shared canonical skill script did not include the Codex fallback, so the earlier fix was effectively overwritten

  In other words, the previous fix was applied to a distributed copy, but not preserved in the canonical shared script that later syncs were based on.

  ## What this PR changes

  - Restores the Codex fallback behavior in the canonical skill-distributed `session-catchup.py`
  - Syncs the updated script back to the generated IDE copies, including `.codex/.../session-catchup.py`
  - Adds a regression test to prevent this behavior from being lost again

  ## Expected behavior

  When the script is running from a Codex skill path and `~/.codex/sessions` exists, but the equivalent Claude project session path does not exist, it should:

  - skip catchup explicitly
  - print a clear fallback message

  instead of silently scanning Claude session storage

  ## Verification

  I added and ran a regression test covering the Codex fallback behavior:

  ```bash
  python -m unittest discover -s tests -v

  And verified the generated IDE folders are back in sync:

  python scripts/sync-ide-folders.py --verify

  ## Why this approach

  This PR fixes the problem in the shared source (skills/planning-with-files/scripts/session-catchup.py) rather than patching only the Codex copy again, so future syncs do not overwrite the
  fix a second time.

### DIFF
--- a/.adal/skills/planning-with-files/scripts/session-catchup.py
+++ b/.adal/skills/planning-with-files/scripts/session-catchup.py
@@ -13,18 +13,31 @@ import sys
 import os
 from pathlib import Path
 from typing import List, Dict, Optional, Tuple
-from datetime import datetime
 
 PLANNING_FILES = ['task_plan.md', 'progress.md', 'findings.md']
 
 
-def get_project_dir(project_path: str) -> Path:
-    """Convert project path to Claude's storage path format."""
+def get_project_dir(project_path: str) -> Tuple[Optional[Path], Optional[str]]:
+    """Resolve session storage path for the current runtime variant."""
     sanitized = project_path.replace('/', '-')
     if not sanitized.startswith('-'):
         sanitized = '-' + sanitized
     sanitized = sanitized.replace('_', '-')
-    return Path.home() / '.claude' / 'projects' / sanitized
+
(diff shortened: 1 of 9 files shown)


OthmanAdi:
Merged. You caught that the v2.23.0 sync overwrote the Codex fallback and fixed it in the canonical source instead of patching just the copy again. That means future syncs won't break it. The regression test helps too. You're credited in the v2.24.0 release. Thanks.

PR #111, opened 2026-03-19
feat: add block-no-verify hook to prevent bypassing git hooks

### LINKED ISSUE #110: Add block-no-verify to prevent AI agents from skipping git hooks
The bypass flag allows agents to skip pre-commit hooks. block-no-verify@1.1.2 blocks this. See https://github.com/tupe12334/block-no-verify for installation instructions for Cursor (preToolUse), Mastra Code (PreToolUse), and Gemini (BeforeTool). Disclosure: the author and maintainer of block-no-verify opened this issue.

tupe12334:
Closes #110. Adds block-no-verify@1.1.2 as the first preToolUse entry in .cursor/hooks.json to block the git hook bypass flag before any shell execution. This complements the existing planning skill hooks without replacing them.


OthmanAdi:
Hey Ofek, appreciate the PR and the transparency about maintaining block-no-verify yourself.

The problem is real, agents bypassing pre-commit hooks can let broken code slip through, and most people don't notice until it's too late. So the intent here is solid.

One thing though: the diff shows `hooks.json` collapsed into a single line, which loses the readable formatting and makes future diffs hard to review. If you keep the original multi-line structure and just slot your entry into the `preToolUse` array, the change would be much cleaner.

Also worth noting, `npx --yes` silently fetches and runs a package on every tool call. Users should probably know that's happening. A short note in the README or a comment in the JSON would go a long way.

The hook itself fits naturally, planning-with-files already leans on `preToolUse` for context refresh, so a guard before shell execution makes sense as a companion.

— Ahmad

PR #114, opened 2026-03-24
feat: publish stack handshake contract and planning state adapter

Charpup:
## Summary
Introduce initial stack-handshake contract artifacts for cross-skill interoperability.

## Changes
- add `contracts/stack-handshake.json`
- add `scripts/stack_contract.py`

## Why
This is a minimal, backward-compatible foundation for future stack coupling across planning/workflow/spec-tooling.

## Notes
- Scope intentionally small (2 files, 1 commit)
- No behavioral change to existing default user flow

### DIFF
--- a/contracts/stack-handshake.json
+++ b/contracts/stack-handshake.json
@@ -1,0 +1,15 @@
+{
+  "contractVersion": "1.0.0",
+  "component": "planning-with-files",
+  "interface": {
+    "health": "bash scripts/check-continue.sh",
+    "capabilities": [
+      "task_plan.md management",
+      "findings.md management",
+      "progress.md management",
+      "session-catchup"
+    ],
+    "export_state": "python scripts/stack_contract.py export --output <file>",
+    "import_state": "python scripts/stack_contract.py import --input <file>"
+  }
+}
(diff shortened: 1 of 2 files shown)


OthmanAdi:
Hey Charpup, thanks for putting this together. I can see the thought behind it and the code is clean.

I'm currently studying this PR in detail and thinking through how it fits with the broader direction of the project. I want to give it proper consideration rather than a rushed review.

I'll get back to you here very soon with my thoughts. Sit tight.

OthmanAdi:
Hi @Charpup, thanks for putting this together. The code is clean and the interoperability idea is interesting.

I'm closing this PR though. Let me explain why.

The PR adds a stack handshake contract and export/import utilities so other skills can read planning files. The problem is nobody has actually asked for this. No issue, no user request, no skill that needs it. It's solving a problem we don't have yet.

The bigger concern: once `contracts/stack-handshake.json v1.0.0` lands in main, I own that spec. Contributors will propose more contracts (analytics/, logging/, etc.) and the project drifts from "persistent markdown planning" into "planning tools plus integration hub." The narrow focus is what makes this work.

If you need this for your own skill or an OpenClaw experiment, fork and extend freely. If cross-skill integration becomes something multiple tools actually need, that's better defined at the ecosystem level (something like an OpenClaw spec) rather than inside planning-with-files.

If someone files an issue with a concrete use case for this, I'm happy to revisit.

Thanks for the contribution and for understanding.

PR #119, opened 2026-04-05
fix(codex): make planning scripts executable in .codex skill install

Leon-Algo:
## Summary
- restore executable bits for the direct-run scripts under `.codex/skills/planning-with-files/scripts/`
- add a Codex-specific regression test that checks those scripts remain executable after checkout/install

## Why
The Codex install path copies the skill from `.codex/skills/planning-with-files/`. In `ide/codex`, the installed copies of `init-session.sh`, `check-complete.sh`, and `session-catchup.py` are currently tracked without execute permission, so direct invocation fails with `permission denied`.

This change keeps the fix intentionally narrow: no script logic changes, only the file modes needed for the installed Codex scripts to run directly.

## Validation
- `python3 -m unittest tests.test_codex_script_permissions -v`
- `python3 tests/test_path_fix.py`
- `./.codex/skills/planning-with-files/scripts/check-complete.sh`
- `./.codex/skills/planning-with-files/scripts/session-catchup.py "$(pwd)"`
- `tmpdir=$(mktemp -d) && cd "$tmpdir" && /Users/leon/Documents/CodeProject/planning-with-files/.codex/skills/planning-with-files/scripts/init-session.sh demo`
- `python3 -m unittest tests.test_session_catchup -v` *(still shows the pre-existing `test_codex_variant_uses_claude_path_when_project_exists` failure, unrelated to this permission-only change)*


OthmanAdi:
Hey @Leon-Algo, thanks for catching this. The permission fix is clean and the regression test is a good add.

Quick question: did you specifically target `ide/codex` on purpose, or did you mean for this to go into `master`? I ask because I ran an audit and the same missing executable bits exist across almost every IDE branch and the core `skills/planning-with-files/scripts/` folder on master too. If you're up for it, a broader fix covering the other branches would be welcome.

Either way, this is merged and you're credited in CONTRIBUTORS.md and CHANGELOG for v2.30.1. Appreciate the catch.

Leon-Algo:
Targeting `ide/codex` was intentional for the immediate Codex installer breakage, but your audit note makes sense. I am following up with a focused PR against `master` for the canonical `skills/planning-with-files/scripts/` execute bits so the fix lands at the source first.

PR #145, opened 2026-04-29
fix: use env bash shebangs

Emin017:
Use /usr/bin/env bash so scripts resolve bash through PATH instead of assuming /bin/bash exists. This improves shebang compatibility on systems such as NixOS.


OthmanAdi:
Thanks @Emin017. Good catch; systems like NixOS that don't have bash at `/bin/bash` will pick this up automatically now.

Issue #148, opened 2026-05-01
Parallel multi-task workflow feedback on v2.0.0+ hooks


shawnli1874:
Sharing some feedback from a parallel multi-task / multi-agent setup, in case it's useful.

### Our scenario

We often have several coding agents working on the same repo at once — for example one terminal running an incident investigation, another implementing a feature, another doing a refactor in the background. Multi-agent parallel work is a common pattern for us across projects.

### How we configured it on v1.x

When the skill was a SKILL.md guide without hooks, we had a line in our project's `CLAUDE.md`:

> When using planning-with-files in this repo, place planning files under `docs/tasks/<task-name>/` rather than the project root.

The LLM read both the skill and our project guide and created files in the chosen folder. Each parallel task got its own subdirectory by convention, and parallel sessions stayed isolated.

### What changed in v2.0.0

v2.0.0 added the `PreToolUse` and `Stop` hooks (CHANGELOG):

> - `PreToolUse` hook: Automatically reads `task_plan.md` before Write/Edit/Bash operations
> - `Stop` hook: Verifies all phases are complete before stopping

The shell commands inside those hooks reference `task_plan.md` as a bare relative path against `cwd`. There doesn't appear to be a path parameter or environment override, so the hooks always look at `<cwd>/task_plan.md` regardless of any project-level convention.

In our setup, the practical effect is:

- `init-session.sh` writes the three files to the project root
- The LLM no longer follows our `CLAUDE.md` placement convention, because the hooks make a different decision
- Multiple parallel sessions all read and write the same three root-level files

After a few hours of parallel work, the files end up looking like (anonymized):

```
# task_plan.md
# Task Plan: Backend Refactor
## Phase 1 [complete]
...

# progress.md
## 2026-01-10 — Backend refactor research
## 2026-01-10 — Incident root cause investigation
## 2026-01-11 — Feature X spec
## 2026-01-11 — Backend refactor POC plan
...
```

Symptoms we've seen:

- Phase status set by one task occasionally gets touched by another
- The "where am I?" reboot test sometimes returns the wrong context
- Findings from one task end up in another task's research notes
- The `head -30 task_plan.md` that runs on every Write/Edit/Bash can show a different task's goals

The escape hatch we previously had via project-level `CLAUDE.md` no longer applies once the hooks are in play.

### Current state

[...]

PR #115, opened 2026-03-24
feat(templates): add analytics workflow template

mvanhorn:
## Summary

Adds an analytics-specific template set for data exploration workflows, plus a `--template` flag on `init-session.sh` / `init-session.ps1` to select it.

## Why

The existing templates assume a software dev workflow (Requirements, Planning, Implementation, Testing, Delivery). Analytics sessions have different phases and tracking needs. From your comment on #103:

> "Things like 'what data sources did I connect to,' 'what did the distribution look like,' 'which features mattered' need their own structure."

| Source | Signal |
|--------|--------|
| [#103](https://github.com/OthmanAdi/planning-with-files/issues/103) | Maintainer requested template ideas for analytics workflows |
| [#19](https://github.com/OthmanAdi/planning-with-files/issues/19) | Users asking how to adapt the skill for non-dev workflows |

## Changes

**New files:**
- `templates/analytics_task_plan.md` - 4-phase plan: Data Discovery, Exploratory Analysis, Hypothesis Testing, Synthesis & Reporting. Replaces "Key Questions" with a "Hypotheses" section. Uses the same `**Status:** pending/in_progress/complete` format so `check-complete.sh` works without modification.
- `templates/analytics_findings.md` - Replaces generic sections with: Data Sources table (source, location, size, key fields, quality notes), Hypothesis Log (hypothesis, test method, result, confidence), Query Results, and Statistical Findings (test, p-value, effect size, conclusion). Keeps Visual/Browser Findings for chart screenshots.

**Modified files:**
- `scripts/init-session.sh` - Added `--template` / `-t` flag. When `--template analytics` is passed, copies analytics templates. Default behavior unchanged. Also generates an analytics-specific `progress.md` with a Query Log table instead of Test Results.
- `scripts/init-session.ps1` - Same `-Template` parameter for Windows parity.

## Testing

Default path (no flag):
```
$ ./init-session.sh "test-project"
Initializing planning files for: test-project (template: default)
Created task_plan.md   # 5 phases, identical to current behavior
```

Analytics path:
```
$ ./init-session.sh --template analytics "user-churn-analysis"
Initializing planning files for: user-churn-analysis (template: analytics)
Created task_plan.md   # 4 phases: Data Discovery, Exploratory, Hypothesis, Synthesis
Created findings.md    # Data Sources, Hypothesis Log, Query Results, Statistical Findings
Created progress.md    # Query Log instead of Test Results
```

`check-complete.sh` compatibility:
```
[...]

PR #149, opened 2026-05-01
fix(scripts): propagate v2.36.0 slug-mode init-session to canonical skill copy + IDE folders

voidborne-d:
## Bug

The v2.36.0 release notes promise:

> **init-session.sh** now accepts a task name and creates a dated, readable plan directory under `.planning/YYYY-MM-DD-<slug>/`. (#148, @shawnli1874)

That rewrite landed in `scripts/init-session.sh` (top-level) but never reached the canonical `skills/planning-with-files/scripts/init-session.{sh,ps1}` that ships in the plugin. Users installing the skill via any of the nine IDE folders (`.codex/`, `.gemini/`, `.pi/`, `.codebuddy/`, `.continue/`, `.factory/`, …) get the v2.0.0 legacy-only init-session and **silently miss** the parallel-plan isolation feature called out as the v2.36.0 headline.

The same gap quietly existed for v2.29.0: the `--template analytics` flag from #115 (@mvanhorn) was added only to the top-level scripts and never to the canonical skill copy.

```
$ grep -c 'SLUG_MODE\|--plan-dir\|slugify' scripts/init-session.sh
12
$ grep -c 'SLUG_MODE\|--plan-dir\|slugify' skills/planning-with-files/scripts/init-session.sh
0
```

## Fix

- **Sync canonical to match top-level**: byte-for-byte copy of `scripts/init-session.{sh,ps1}` into `skills/planning-with-files/scripts/`. Brings slug mode, `--plan-dir`, `--template analytics`, and the v2.36.0 Windows-aware `short_uuid` probing into the shipped skill.
- **Refresh IDE mirrors via the existing tool**: `python scripts/sync-ide-folders.py` propagates the canonical to `.codebuddy/`, `.codex/`, `.continue/`, `.factory/`, `.gemini/`, `.pi/`. Same pass picks up the v2.35.1 `/usr/bin/env bash` shebang on `check-complete.sh` in those IDE folders — Emin017's #145 landed against the canonical only and the IDE mirrors had drifted.
- **Add a regression test** so this can't silently happen again:
  - `tests/test_canonical_script_sync.py::test_shared_scripts_match_canonical_copy` — `init-session.{sh,ps1}` and `check-complete.{sh,ps1}` must be byte-identical between `scripts/` and `skills/planning-with-files/scripts/`. `session-catchup.py` is **intentionally excluded** from this check (the canonical copy carries Codex-specific catchup logic that doesn't live at the top level).
  - `tests/test_canonical_script_sync.py::test_sync_ide_folders_verify_clean` — invokes `python scripts/sync-ide-folders.py --verify` and asserts exit 0, so any future commit that updates the canonical without re-running the sync trips the test in CI.

## Scope discipline

- The five language variants (`planning-with-files-{ar,de,es,zh,zht}`) have separate copies that are not auto-synced and also lack slug-mode. Out of scope here — they're not in `IDE_MANIFESTS` and would deserve a translator-aware follow-up. Happy to send a second PR if you want them lined up.
- `session-catchup.py` is intentionally not part of the parity assertion. The canonical version has Codex catchup features (orjson, MIN_SESSION_BYTES, current-thread preference) that the top-level version does not. If you want them merged that's its own conversation; this PR doesn't touch the question.

## Local gates

- `python3.11 -m unittest discover tests` → **70 pass / 1 pre-existing fail**. The single fail is `test_installed_plugin_resolves_repo_assets_for_completion_check` in `test_hermes_adapter.py`, an unrelated macOS `/private/var/...` tempdir-symlink path comparison that already fails on upstream master at f7e8919 (zero changes from this branch). New tests pass: 70 = 68 baseline + 2 new.
- `python3.11 -m unittest tests/test_init_session_slug.py` → 6/6 pass. The pre-existing slug-mode behavior tests still target the top-level script and still pass; the parity assertion in the new test ensures the canonical copy now exhibits the same behavior.
- **End-to-end smoke** of the canonical skill copy:

```
$ bash skills/planning-with-files/scripts/init-session.sh "Smoke Test Run"
Initializing planning files for: Smoke Test Run (template: default)
PLAN_ID=2026-05-02-smoke-test-run
Created /tmp/smoke/.planning/2026-05-02-smoke-test-run/task_plan.md
Created /tmp/smoke/.planning/2026-05-02-smoke-test-run/findings.md
Created /tmp/smoke/.planning/2026-05-02-smoke-test-run/progress.md
Active plan recorded: /tmp/smoke/.planning/.active_plan
```

Legacy mode (zero args) in a separate tmp dir keeps root-level files and doesn't create `.planning/` — backward-compat preserved.

## Why this PR is +1727/-257 across 20 files

The line counts make this look bigger than it is. The actual logic change is:

1. One script (`init-session.sh`) and one PowerShell script (`init-session.ps1`) updated in the canonical location.
2. `sync-ide-folders.py` then mirrors the canonical into 6 IDE folders × 3 files each = 18 mechanical copies.
3. One new 75-line test file.

The non-test diff is essentially the same ~280-line file replicated across IDE folders that the existing sync tool generates.

## Generated with Claude Code

Per the project's recent merge precedent for `🤖 Generated with` trailers (PRs #126, #145, etc.), full disclosure: implementation drafted with Claude Code and verified against the local test suite + smoke-tested before push. All assertions in this PR body were checked locally.

### DIFF
--- a/.codebuddy/skills/planning-with-files/scripts/check-complete.sh
+++ b/.codebuddy/skills/planning-with-files/scripts/check-complete.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # Check if all phases in task_plan.md are complete
 # Always exits 0 — uses stdout for status reporting
 # Used by Stop hook to report task completion status
--- a/.codebuddy/skills/planning-with-files/scripts/init-session.ps1
+++ b/.codebuddy/skills/planning-with-files/scripts/init-session.ps1
@@ -1,17 +1,34 @@
 # Initialize planning files for a new session
-# Usage: .\init-session.ps1 [project-name]
+# Usage: .\init-session.ps1 [-Template TYPE] [project-name]
+# Templates: default, analytics
 
 param(
-    [string]$ProjectName = "project"
+    [string]$ProjectName = "project",
+    [string]$Template = "default"
 )
 
 $DATE = Get-Date -Format "yyyy-MM-dd"
 
-Write-Host "Initializing planning files for: $ProjectName"
+# Resolve template directory (skill root is one level up from scripts/)
+$ScriptDir = Split-Path -Parent $MyInvocation.MyCommand.Path
+$SkillRoot = Split-Path -Parent $ScriptDir
+$TemplateDir = Join-Path $SkillRoot "templates"
+
+Write-Host "Initializing planning files for: $ProjectName (template: $Template)"
+
+# Validate template
+if ($Template -ne "default" -and $Template -ne "analytics") {
+    Write-Host "Unknown template: $Template (available: default, analytics). Using default."
+    $Template = "default"
+}
 
 # Create task_plan.md if it doesn't exist
 if (-not (Test-Path "task_plan.md")) {
-    @"
+    $AnalyticsPlan = Join-Path $TemplateDir "analytics_task_plan.md"
+    if ($Template -eq "analytics" -and (Test-Path $AnalyticsPlan)) {
+        Copy-Item $AnalyticsPlan "task_plan.md"
+    } else {
+        @"
 # Task Plan: [Brief Description]
 
 ## Goal
@@ -56,6 +73,7 @@ Phase 1
 | Error | Resolution |
 |-------|------------|
 "@ | Out-File -FilePath "task_plan.md" -Encoding UTF8
+    }
     Write-Host "Created task_plan.md"
 } else {
     Write-Host "task_plan.md already exists, skipping"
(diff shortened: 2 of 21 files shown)


OthmanAdi:
@voidborne-d — merged in v2.36.2.

The bug was exactly as you described. `skills/planning-with-files/scripts/init-session.sh` had zero slug-mode code; anyone installing via `npx skills add` or any of the IDE folders got the v2.0.0 script and silently missed the v2.36.0 headline feature. The analytics template and shebang gaps you spotted alongside it were real too.

The regression test is the part that matters most here. The byte-comparison assertion plus the `--verify` check means this class of drift will trip CI going forward rather than shipping quietly.

You're in CONTRIBUTORS.md.

PR #167, opened 2026-05-22
Skip script permission test on Windows

gauravvojha:
## Summary

Skip executable permission checks on Windows because NTFS does not preserve POSIX executable bits in the same way as Linux/macOS.

## Changes

* Added `pytest.mark.skipif` for Windows platform in `tests/test_script_permissions.py`
* Added clear reason message explaining the filesystem limitation

## Testing

Tested on:

* Windows 11
* Python 3.11

Command used:

```bash
python -m pytest tests/test_script_permissions.py -v
```

Result:

* Tests skipped on Windows as expected

### DIFF
--- a/tests/test_script_permissions.py
+++ b/tests/test_script_permissions.py
@@ -1,3 +1,5 @@
+import sys
+import pytest
 import stat
 import unittest
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


OthmanAdi:
@gauravvojha fix confirmed in v2.41.0. Root cause: NTFS does not preserve POSIX executable bits, so the two exec-bit tests in test_script_permissions.py always reported mode 0o100666 on Windows. Mechanism: a class-level pytest.mark.skipif decorator now skips the tests on sys.platform == 'win32' while leaving them active on Linux and macOS. Your PR's patch had a duplicate standalone function and a nested method inside it at merge time, so the skipif was re-applied at class-level granularity in the post-merge repair. Test logic is otherwise identical. You are listed in CONTRIBUTORS.md.

PR #180, opened 2026-05-30
fix: make Codex stop hook non-blocking for incomplete plans

### LINKED ISSUE #178: Codex Stop hook blocks normal stopping and causes the agent to continue unfinished plans automatically
## Summary

The Codex Stop hook currently treats an unfinished `planning-with-files` plan as a condition that should block stopping. This can cause the agent to continue into the next phase without an explicit user request.

This is risky because an incomplete plan is a normal state. The hook should remind the agent/user to sync progress, but it should not force continuation or instruct the agent to keep working on remaining phases.

## Environment

- Tool: Codex
- Skill: `planning-with-files`
- Skill version: `2.43.0`
- Project has scoped planning files under `.planning/<plan-id>/`
- `.planning/.active_plan` points to the active plan directory
- Plan was partially complete, for example `6/12` phases complete

## Reproduction

1. Install or enable the Codex hooks from `planning-with-files`.
2. In a project, create a scoped plan:
   - `.planning/.active_plan`
   - `.planning/<plan-id>/task_plan.md`
   - `.planning/<plan-id>/progress.md`
   - `.planning/<plan-id>/findings.md`
3. Make `task_plan.md` contain multiple `### Phase` sections, with some phases still pending or in progress.
4. Ask the agent to do a bounded task, such as only updating progress or only answering a question.
5. When the agent tries to stop, the Stop hook reports:

```text
[planning-with-files] Task incomplete (6/12 phases done). Update progress.md, then read task_plan.md and continue working on the remaining phases.
```

6. The agent then continues working on the next phase automatically, even though the user did not ask it to continue.

## Actual behavior

The Stop hook emits a blocking decision for incomplete plans.

In the Codex hook path, `stop.py` turns the incomplete-plan message into:

```json
{"decision": "block", "reason": "...continue working on the remaining phases."}
```

This makes the agent treat the hook output as a system-level instruction to keep working. As a result, it may advance the plan and edit `task_plan.md`, `progress.md`, or `findings.md` unexpectedly.

## Expected behavior

An incomplete plan should be treated as a normal state.

The Stop hook should not block stopping just because `COMPLETE < TOTAL`. It should emit a non-blocking status message such as:

```text
[planning-with-files] Task in progress (6/12 phases complete). If ending this turn, make sure progress.md is up to date.
```

It should not include imperative continuation text like:

```text
continue working on the remaining phases
```

The agent should only continue to the next phase when the user explicitly asks it to continue.

## Suspected root cause

There are two related issues in the Codex hook implementation:

1. `.codex/hooks/stop.py` emits `decision: block` for incomplete plans.
2. `.codex/hooks/stop.sh` uses command-like wording:

```text
Update progress.md, then read task_plan.md and continue working on the remaining phases.
```

This combination converts a status check into a forced continuation command.

There is also a confusing mismatch: `skills/planning-with-files/scripts/check-complete.sh` already has safer wording:

```text
Task in progress (...). Update progress.md before stopping.
```

But the Codex hook path uses `.codex/hooks/stop.py -> .codex/hooks/stop.sh`, so updating the skill script alone does not fix the behavior.

## Suggested fix

- Do not emit `decision: block` for incomplete plans.
- Emit a non-blocking `systemMessage` or equivalent status notification instead.
- Remove `continue working on the remaining phases` from the Stop hook message.
- Only block stopping for actual safety problems, such as plan tampering or invalid hook state, not for normal incomplete work.

## Why this matters

The current behavior can make the agent perform work outside the user's explicit request. For planning tools, this is especially problematic because it can mark phases complete or move the active plan forward without user intent.

2023Anita:
## Summary

- make the Codex Stop hook report incomplete plans as a non-blocking `systemMessage`
- remove the imperative "continue working on the remaining phases" wording from the stop message
- update the Codex hook test to assert that incomplete plans do not emit `decision: block`

## Motivation

Closes #178.

An incomplete plan is a normal planning state. Blocking Stop for `COMPLETE < TOTAL` can make Codex continue into another phase without an explicit user request. This keeps the handoff reminder, but avoids turning a status check into a continuation command.

## Verification

```bash
python3 -m unittest tests.test_codex_hooks tests.test_codex_session_isolation
```

Passed locally.

I also ran the full unittest discovery. It reached unrelated existing environment/repo issues: missing local `pytest`, macOS `/private/var` temp path normalization, and missing `clawhub-upload/SKILL.md`. The targeted Codex hook tests above pass.

### DIFF
--- a/.codex/hooks/stop.py
+++ b/.codex/hooks/stop.py
@@ -18,15 +18,7 @@ def main() -> None:
     if not isinstance(message, str) or not message:
         return
 
-    if "ALL PHASES COMPLETE" in message:
-        adapter.emit_json({"systemMessage": message})
-        return
-
-    if bool(payload.get("stop_hook_active")):
-        adapter.emit_json({"systemMessage": message})
-        return
-
-    adapter.emit_json({"decision": "block", "reason": message})
+    adapter.emit_json({"systemMessage": message})
 
 
 if __name__ == "__main__":
--- a/.codex/hooks/stop.sh
+++ b/.codex/hooks/stop.sh
@@ -30,5 +30,5 @@ if [ "$COMPLETE" -eq "$TOTAL" ] && [ "$TOTAL" -gt 0 ]; then
     exit 0
 fi
 
-echo "{\"followup_message\": \"[planning-with-files] Task incomplete ($COMPLETE/$TOTAL phases done). Update progress.md, then read task_plan.md and continue working on the remaining phases.\"}"
+echo "{\"followup_message\": \"[planning-with-files] Task in progress ($COMPLETE/$TOTAL phases complete). If ending this turn, make sure progress.md is up to date.\"}"
 exit 0
--- a/tests/test_codex_hooks.py
+++ b/tests/test_codex_hooks.py
@@ -147,7 +147,7 @@ class CodexHooksTests(unittest.TestCase):
         payload = json.loads(result.stdout)
         self.assertIn("progress.md", payload["systemMessage"])
 
-    def test_stop_adapter_blocks_once_then_allows_reentry(self) -> None:
+    def test_stop_adapter_reports_incomplete_plan_without_blocking(self) -> None:
         with tempfile.TemporaryDirectory() as tmpdir:
             root = Path(tmpdir)
             root.joinpath("task_plan.md").write_text(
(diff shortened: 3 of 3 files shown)


instinct6819-oss (approved):

instinct6819-oss (approved):

OthmanAdi:
Merged, @2023Anita. This is the right fix.

`.codex/hooks/stop.py` was emitting `{"decision": "block"}` on the first stop while phases were still pending, and that is what pushed the agent to keep going on its own. Your change drops that path and keeps an advisory reminder only, which is exactly the v3 line that an incomplete plan alone should never block a stop. It shipped in v3.1.0, your commit is on master with your authorship intact, and you are credited in CONTRIBUTORS.md and the CHANGELOG. This closes #178. Thank you.

PR #182, opened 2026-06-11
fix: add version, license, and metadata.hermes.tags to all 17 skills

shunfeng8421:
- All 17 SKILL.md files were missing version field
- 16/17 were missing license field
- All 17 were missing metadata.hermes.tags (Hermes convention)

Includes multilingual variants (Arabic, German, Spanish, Chinese).

Audited by Hermes Skill Auditor. Before: 60/100 (C)


OthmanAdi:
Thanks for running the audit and opening this, @shunfeng8421. I read the diff closely, and I am going to pass on this one. Let me be specific about why, since a vague no is no use to you.

Two reasons. The version field is already present: v3.0.0 added `metadata.version: "3.0.0"` to the canonical SKILL.md and every parity-locked variant on 2026-06-10, so the "missing version" premise does not hold against current master. The blocking reason is the shape of the patch. It appends a second `metadata:` key under the existing one in each file, and YAML does not allow duplicate mapping keys. A compliant parser keeps only the last `metadata:` block, so `metadata.version` quietly disappears from the parsed frontmatter. The regex-based parity test would still pass, which means this would land as a silent corruption across all 17 files rather than an improvement. The diff also edits `.continue` and `.gemini`, which we hold back from the parity set on purpose, and gives `.pi` a `1.0.0` that does not match its scheme.

If a Hermes `tags` convention is worth adopting, the right shape is a single nested key under the existing `metadata:` block, with a link to the Hermes spec it follows, since the field has no precedent in the repo yet. I am happy to look at that as a separate, smaller PR. Appreciate you engaging with the project.

PR #183, opened 2026-06-14
fix: sync legacy skills/ SKILL.md to IDE-specific paths

xwang118:
## What

The legacy `skills/planning-with-files/SKILL.md` still referenced `${CLAUDE_PLUGIN_ROOT}` in its prose body, while the plugin copy `planning-with-files/SKILL.md` was updated to explicit per-IDE paths in 7de9247. This syncs the legacy copy so the two are identical.

## Why

Non-Claude-Code installs read the legacy `skills/` copy (Codex `~/.codex/...`, Cursor `~/.cursor/...`). With the old `${CLAUDE_PLUGIN_ROOT}` references, those users got paths that don't resolve in their IDE. Two spots were stale:

- the session-catchup command block
- the "Important: Where Files Go" section

## Changes

- `skills/planning-with-files/SKILL.md`: session-catchup block now lists Claude Code / Codex / Cursor variants; "Where Files Go" uses IDE-specific template paths — now byte-identical to the plugin copy.

Hooks frontmatter still uses `${CLAUDE_PLUGIN_ROOT}` (correct — that resolves in the plugin runtime); only the human-readable prose paths changed. No functional/runtime change for Claude Code users.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/skills/planning-with-files/SKILL.md
+++ b/skills/planning-with-files/SKILL.md
@@ -43,7 +43,14 @@ Work like Manus: Use persistent markdown files as your "working memory on disk."
 **Before starting work**, check for unsynced context from a previous session:
 
 ```bash
-python3 ${CLAUDE_PLUGIN_ROOT}/scripts/session-catchup.py "$(pwd)"
+# Claude Code users
+python3 ~/.claude/skills/planning-with-files/scripts/session-catchup.py "$(pwd)"
+
+# Codex users
+python3 ~/.codex/skills/planning-with-files/scripts/session-catchup.py "$(pwd)"
+
+# Cursor users
+python3 ~/.cursor/skills/planning-with-files/scripts/session-catchup.py "$(pwd)"
 ```
 
 If catchup report shows unsynced context:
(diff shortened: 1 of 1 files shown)


OthmanAdi:
You spotted something real: `CLAUDE_PLUGIN_ROOT` does not expand when a skill-only user pastes the documented session-catchup command into their own shell, so that one command is broken for non-plugin installs. That part is worth fixing.

This particular change cannot merge, for three reasons.

1. The file it syncs to, `planning-with-files/SKILL.md`, no longer exists. It was the old duplicate plugin copy and was removed in `3e08dd9` (`remove duplicate planning-with-files folder, only skills should exist`). There is no second copy left to match, so byte-for-byte alignment has no target.

2. `skills/planning-with-files/SKILL.md` keeps `CLAUDE_PLUGIN_ROOT` on purpose. It resolves in the plugin runtime and stays correct across the 60-plus supported adapters. Replacing it with three hardcoded home paths (Claude, Codex, Cursor) covers three of those adapters and breaks the plugin install itself, which lands under `~/.claude/plugins/`, not `~/.claude/skills/`.

3. The diff also deletes the Windows PowerShell session-catchup block. Windows users rely on it, and all five language variants carry it.

The narrow fix (making the documented command work for skill-only installs) is tracked in #185. If you want it, it is yours. Scope: keep `CLAUDE_PLUGIN_ROOT` for the plugin route, keep the Windows block, and fix only the copy-paste command for the non-plugin case.

Closing this PR.

PR #186, opened 2026-06-15
fix: session-catchup command fails when PLUGIN_ROOT is unset (fixes #185)

### LINKED ISSUE #185: Session-catchup command fails outside the plugin install route because CLAUDE_PLUGIN_ROOT is unset
## Problem

The documented Linux/macOS session-catchup command in `skills/planning-with-files/SKILL.md` relies on `${CLAUDE_PLUGIN_ROOT}`:

```bash
$(command -v python3 || command -v python) ${CLAUDE_PLUGIN_ROOT}/scripts/session-catchup.py "$(pwd)"
```

`CLAUDE_PLUGIN_ROOT` is an environment variable the Claude Code plugin runtime sets only when it runs hook commands. It is not present in a normal interactive shell. Users who installed the skill outside the plugin route (via `npx skills add`, or on Codex or Cursor) and copy-paste this command into a terminal get an empty variable. The command collapses to `python3 /scripts/session-catchup.py "$(pwd)"`, which points at an absolute `/scripts` path that does not exist, and the command fails.

## Why it happens and how to reproduce

`${CLAUDE_PLUGIN_ROOT}` resolves correctly inside a plugin-executed hook, so the plugin install route works. In an interactive shell the variable is undefined, so shell expansion produces a leading slash and a wrong path.

Repro (any shell where the plugin runtime is not the caller):

```bash
echo "${CLAUDE_PLUGIN_ROOT}/scripts/session-catchup.py"
# prints: /scripts/session-catchup.py
```

Where the skill actually lands depends on the install route:

- Plugin install (`/plugin install`): under `~/.claude/plugins/`.
- Skill-only install: under `~/.claude/skills/planning-with-files/`.
- `npx skills add`: via a junction into the project's `.agents/skills/`.

So the catchup script lives in different places per route, and the single `${CLAUDE_PLUGIN_ROOT}` reference only resolves for one of them.

## Constraints

A fix here should avoid the regressions that kept PR #183 from landing:

- Keep `${CLAUDE_PLUGIN_ROOT}` working for the plugin install route, where it resolves correctly. Do not drop the variable for plugin users.
- Keep the Windows PowerShell catchup block intact.
- Do not desync the five language variants (`ar`, `de`, `es`, `zh`, `zht`). Mirror whatever lands in the canonical file.
- Be honest about where the skill lands per route (the three cases above). Do not hardcode a single IDE path.

Suggested direction, left open for the implementer: offer a route-aware path, or a short resolver note, so skill-only users get a command that runs, without hardcoding one IDE path and without removing the variable for plugin users.

## Credit

@xwang118 spotted this in PR #183. The broader change there could not be merged (it synced to a since-deleted file, dropped the Windows PowerShell block, and hardcoded only three IDE paths), but the underlying observation holds and is worth fixing on its own. @xwang118, if you want to take this scoped fix, it is yours.

shunfeng8421:
## Problem

Fixes #185 — the session-catchup command documented in SKILL.md fails when run outside the plugin runtime because `${CLAUDE_PLUGIN_ROOT}` is only set by the plugin loader.

In an interactive shell, the command collapses to an invalid absolute path.

## Fix

Use shell parameter expansion `${VAR:-fallback}` so the command works in both contexts:

```bash
SKILL_DIR="${CLAUDE_PLUGIN_ROOT:-$HOME/.claude/skills/planning-with-files}"
$(command -v python3 || command -v python) "${SKILL_DIR}/scripts/session-catchup.py" "$(pwd)"
```

- **Plugin users**: `CLAUDE_PLUGIN_ROOT` is set → takes priority, no behavior change
- **Non-plugin users**: falls back to `~/.claude/skills/planning-with-files/` → command works

## Files changed (7 files, +16/-9 lines)

| File | Change |
|------|--------|
| `skills/planning-with-files/SKILL.md` | Canonical — `${CLAUDE_PLUGIN_ROOT}` → fallback |
| `.codebuddy/skills/planning-with-files/SKILL.md` | CodeBuddy — `${CODEBUDDY_PLUGIN_ROOT}` → fallback |
| `skills/planning-with-files-{ar,de,es,zh,zht}/SKILL.md` | 5 language variants — same fix |

Tool-specific variants (`.codex`, `.cursor`, `.kiro`, `.hermes`, etc.) already use hardcoded paths and were not changed.

## Constraints respected

- ✅ `CLAUDE_PLUGIN_ROOT` still works for plugin users (takes priority when set)
- ✅ Windows PowerShell catchup block unchanged
- ✅ Language variants kept in sync
- ✅ No desync with tool-specific variants

### DIFF
--- a/.codebuddy/skills/planning-with-files/SKILL.md
+++ b/.codebuddy/skills/planning-with-files/SKILL.md
@@ -41,8 +41,9 @@ Work like Manus: Use persistent markdown files as your "working memory on disk."
 **Before starting work**, check for unsynced context from a previous session:
 
 ```bash
-# Linux/macOS (auto-detects python3 or python)
-$(command -v python3 || command -v python) ${CODEBUDDY_PLUGIN_ROOT}/scripts/session-catchup.py "$(pwd)"
+# Linux/macOS — auto-detects skill directory (plugin env or default install path)
+SKILL_DIR="${CODEBUDDY_PLUGIN_ROOT:-$HOME/.codebuddy/skills/planning-with-files}"
+$(command -v python3 || command -v python) "${SKILL_DIR}/scripts/session-catchup.py" "$(pwd)"
 ```
 
 ```powershell
--- a/skills/planning-with-files-ar/SKILL.md
+++ b/skills/planning-with-files-ar/SKILL.md
@@ -44,7 +44,8 @@ metadata:
 
 ```bash
 # Linux/macOS
-$(command -v python3 || command -v python) ${CLAUDE_PLUGIN_ROOT}/scripts/session-catchup.py "$(pwd)"
+SKILL_DIR="${CLAUDE_PLUGIN_ROOT:-$HOME/.claude/skills/planning-with-files-ar}"
+$(command -v python3 || command -v python) "${SKILL_DIR}/scripts/session-catchup.py" "$(pwd)"
 ```
 
 ```powershell
--- a/skills/planning-with-files-de/SKILL.md
+++ b/skills/planning-with-files-de/SKILL.md
@@ -44,7 +44,8 @@ Arbeite wie Manus: Verwende persistente Markdown-Dateien als deinen „Festplatt
 
 ```bash
 # Linux/macOS
-$(command -v python3 || command -v python) ${CLAUDE_PLUGIN_ROOT}/scripts/session-catchup.py "$(pwd)"
+SKILL_DIR="${CLAUDE_PLUGIN_ROOT:-$HOME/.claude/skills/planning-with-files-de}"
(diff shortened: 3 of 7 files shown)


OthmanAdi:
Merged and shipped in v3.1.2, adopted by cherry-pick as `bbedaad` with your authorship preserved.

The `${CLAUDE_PLUGIN_ROOT:-$HOME/.claude/skills/planning-with-files}` fallback is the right call: plugin users keep the variable as priority, skill-only installs get a working path, and the Windows PowerShell block stays untouched. I extended the same fix to the `.hermes` variant in the release commit, which carried the identical unguarded `$HERMES_HOME` in both its bash and PowerShell blocks.

You are credited in CONTRIBUTORS.md.

Issue #197, opened 2026-07-08
CI: no workflow runs the pytest or Pi extension test suites


Yigtwxx:
## Problem

The repository has a substantial test suite (202 pytest tests plus 21 vitest tests for the Pi extension), and CONTRIBUTING.md asks contributors to run `python -m pytest tests/ -q` before opening a PR. But no GitHub Actions workflow runs either suite. The only CI today is the Tessl skill review, which triggers on SKILL.md changes and reviews skill prose, not behavior.

This gap matters more here than in most repos:

- Windows regressions are a recurring bug class. In v3.2.0 alone, `session-catchup.py` turned out to be silently non-functional on Windows and `inject-plan.sh`'s containment guard silently dropped injection under 8.3 short-name paths. A `windows-latest` CI leg catches this class at PR time instead of at release audit time.
- The suite itself enforces the repo's core invariants: version parity across the 19 bump targets (`test_skill_md_version_parity.py`) and canonical-vs-mirror script sync (`test_canonical_script_sync.py`, which runs `scripts/sync-ide-folders.py --verify`). A docs-only or variant-only PR can break these invariants and nothing currently checks them automatically.

## Evidence

I ran the suite on GitHub-hosted runners from my fork before proposing this. The very first run exposed two latent test-portability failures, which is a good demonstration of the gap. Neither is a production bug:

1. `tests/test_containment.py::test_active_plan_symlink_out_of_root_is_rejected` fails on `windows-latest`. The hosted runner can create symlinks (unlike most local Windows setups, where the test honestly skips), so the test actually runs there. It then fails because the resolver's stdout is a POSIX-style Git Bash path (`/tmp/...`, the mounted user temp dir), while the assertion canonicalizes it with `os.path.realpath`, which reads it as `C:\tmp`. The resolver rejects the escaping symlink correctly; only the comparison side is wrong. Same class as the normalization fix in 0d028d4.
2. `tests/test_path_fix.py::test_windows_native_path` and `test_git_bash_path` fail on `ubuntu-latest`. `Path.resolve()` treats `C:/Users/...` as a relative path on POSIX and prepends the runner's CWD before sanitization. Production is unaffected: on POSIX hosts the sanitizer only ever receives real POSIX paths, so the branch under test is Windows-only.

Failing run (before fixes): https://github.com/Yigtwxx/planning-with-files/actions/runs/28927948152
[...]

PR #198, opened 2026-07-08
fix: make containment and path-sanitize tests portable to CI runners

Yigtwxx:
## What changed

Two test-side portability fixes, no production script changes:

1. `tests/test_containment.py`: added a `host_realpath()` helper that maps the resolver's sh-emitted POSIX-style path back to its Windows form with `cygpath` before canonicalizing, and used it on the comparison side of the containment assertion.
2. `tests/test_path_fix.py`: restricted `test_windows_native_path` and `test_git_bash_path` to Windows with `skipUnless(os.name == "nt")`, mirroring the existing `skipIf` on the Unix vector in the same file.

## Why

Running the suite on GitHub-hosted runners for the first time (see #197) exposed both as latent failures:

- On `windows-latest` the runner can create symlinks, unlike most local Windows setups, so the escaping-symlink containment test actually runs there. It failed because the resolver prints a Git Bash POSIX path (`/tmp/...`, the mounted user temp dir) while the assertion canonicalized it with `os.path.realpath`, which reads it as `C:\tmp`. The resolver rejected the escape correctly; only the comparison was wrong. This is the same comparison-side normalization class as 0d028d4.
- On `ubuntu-latest` the two Windows-shaped sanitizer vectors failed because `Path.resolve()` treats `C:/Users/...` as a relative path on POSIX and prepends the CWD. On POSIX hosts the sanitizer only ever receives real POSIX paths in production, so the branch under test is Windows-only and the honest fix is to run those vectors only where their input shape is meaningful.

## How I tested it

- Local Windows 11 (symlink-capable, so the containment test runs rather than skips): `python -m pytest tests/ -q` goes from `1 failed, 201 passed, 3 skipped` to `202 passed, 3 skipped`.
- GitHub-hosted runners via my fork: failing run before the fixes https://github.com/Yigtwxx/planning-with-files/actions/runs/28927948152, green run after them https://github.com/Yigtwxx/planning-with-files/actions/runs/28928254774 (both `ubuntu-latest` and `windows-latest`).

## Limitations / notes

- `host_realpath()` falls back to the raw path when `cygpath` is absent; with Git Bash providing `sh` (the guard for the whole test class) `cygpath` is present in practice.
- The two skipped vectors keep full coverage on Windows, which is the platform the guarded regression (v3.2.0 sanitization fix) targets.

Addresses #197 (part 1 of 2; the CI workflow PR builds on this to land green).

### DIFF
--- a/tests/test_containment.py
+++ b/tests/test_containment.py
@@ -32,6 +32,29 @@ def have_sh() -> bool:
     return shutil.which("sh") is not None
 
 
+def host_realpath(sh_path: str) -> str:
+    """Canonicalize a path emitted by an sh script for host-side comparison.
+
+    Under Git Bash on Windows the resolver prints POSIX-style paths, and the
+    user temp directory is mounted at /tmp, so os.path.realpath on the raw
+    string would resolve against a nonexistent C:\\tmp and the containment
+    assertion would fail even though the resolved dir is contained. Map the
+    path back to its Windows form with cygpath first, then canonicalize.
+    """
+    if os.name == "nt" and sh_path.startswith("/"):
+        cygpath = shutil.which("cygpath")
+        if cygpath:
+            mapped = subprocess.run(
+                [cygpath, "-w", sh_path],
+                text=True,
+                capture_output=True,
+                check=False,
+            ).stdout.strip()
+            if mapped:
+                sh_path = mapped
+    return os.path.realpath(sh_path)
+
+
 def can_make_escaping_symlink() -> bool:
     """True only if we can create a dir symlink that realpath sees escaping root.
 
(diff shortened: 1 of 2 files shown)


OthmanAdi:
Merged and in master. `host_realpath()` with the `cygpath` map is the right call: the resolver was rejecting the escaping symlink correctly, only the comparison side read `/tmp` as `C:\tmp`. Same normalization class as 0d028d4. The `skipUnless` on the two Windows-shaped vectors is also correct, since POSIX `Path.resolve()` treats `C:/...` as relative and prepends the CWD. Credited in CONTRIBUTORS.md and the CHANGELOG. Thanks for keeping it test-side only, no production script touched.

Yigtwxx:
Thanks for the quick review and merge. Glad the host_realpath() approach landed cleanly; keeping the normalization on the comparison side only felt like the safest way to preserve what the resolver was already doing right. The skipUnless markers should also keep those two vectors meaningful if the suite ever runs on a POSIX-only matrix leg.

Issue #19, opened 2026-01-12
For multi-step / complex tasks, how should this skill be used properly?


wqh17101:
I am using this skill to handle a complex task. It has planned 5 steps for me. After completing the first step, the context length has already reached **140K**, nearing the automatic compression trigger.
To ensure the best results, how should I proceed? For example, should I open a new window and ask it to continue working by referring to the task_plan.md file? If so, what would be an appropriate prompt to use?

liangxiwei:
just use claude code's plan mode

OthmanAdi:
Good question @wqh17101  this is exactly the scenario the skill is designed for.

**The solution:** Start a new Claude session and use this prompt:

```
Read task_plan.md, findings.md, and progress.md to restore context. Continue from Phase [X].
```

The core idea of this skill is that your **files are your memory**, not the context window. When context fills up, you don't try to compress it  you start fresh and reload from the files.

**Before switching sessions**, make sure your 3 planning files are fully updated. If you've been following the workflow (updating after each phase, logging discoveries), everything important is already on disk.

**Why not use `/resume` or auto-compact?**
- `/resume` restores full message history — same context problem
- Auto-compact summarizes and loses some details
- Fresh session + read files gives you a clean slate with nothing lost

---

**A possible improvement:** I could create a branch with a `RESUME.md` template  a file that gives you a ready-to-paste prompt and captures any volatile context before you switch sessions. If you'd like to test this, let me know and I'll set up a branch for you to try.

OthmanAdi:
@liangxiwei Plan mode solves a different problem it's for designing an approach before implementation.

This issue is about context filling up *during execution* of a multi-step task. Plan mode doesn't help when you're mid-task at 140K tokens.

The two can work together: plan mode for the design phase, planning-with-files for persisting state across sessions during execution.

OthmanAdi:
@wqh17101 中文版本：

好问题 这正是这个技能设计的使用场景。

**解决方案：** 开启一个新的 Claude 会话，使用这个提示词：

```
读取 task_plan.md、findings.md 和 progress.md 来恢复上下文。从第 [X] 阶段继续。
```

这个技能的核心理念是：**文件就是你的记忆**，而不是上下文窗口。当上下文满了，不要试图压缩它 直接开新会话，从文件重新加载。

[...]

Issue #212, opened 2026-08-01
[regression] Agent Skills inline hooks can inject another Codex thread's active plan from a shared cwd


webwww123:
## Summary

In Codex Desktop using the Agent Skills / skill-only install route (`~/.codex/skills/planning-with-files/SKILL.md`), the inline lifecycle hooks can inject an unrelated task's active plan when multiple Codex threads share the same top-level cwd.

This is not plan-file content corruption. Two separate plans remain correct, but the shared `.planning/.active_plan` selector chooses the wrong plan and re-injects it on every prompt/tool call. Because the injected plan arrives as high-priority context, the agent can abandon the user's current task and start working on an unrelated one.

We reproduced this on planning-with-files v3.7.0. I also checked current master/v3.8.2: the inline `skills/planning-with-files/SKILL.md` hook commands still resolve only by `PLAN_ID -> cwd .planning/.active_plan -> newest plan` and do not appear to apply the Codex session-attachment guard.

## Impact

In the observed incident, the user was discussing a package-switching implementation in a nested project. A different root-level plan described an unrelated bulk production migration. The wrong plan was injected repeatedly, and the agent began investigating the unrelated migration.

Independent production-safety checks prevented writes, but this failure mode is high risk: a stale or unrelated plan can redirect an active coding/operations thread without an explicit user task switch.

## Reproduction

1. Install planning-with-files via the Agent Skills / skill-only route so hooks come from `SKILL.md`.
2. Start Codex Desktop with a shared cwd such as `/workspace`.
3. Create Plan A under:
   - `/workspace/.planning/plan-a/task_plan.md`
   - `/workspace/.planning/.active_plan` contains `plan-a`
4. Create an unrelated Plan B in a nested project:
   - `/workspace/project/.planning/plan-b/task_plan.md`
   - `/workspace/project/.planning/.active_plan` contains `plan-b`
5. In a Codex thread, discuss and work on Plan B. Tool calls may explicitly use `workdir=/workspace/project`.
6. Invoke planning-with-files or make tool calls covered by its hooks.

## Actual behavior

The hook resolves against the Codex thread's shared cwd (`/workspace`), selects Plan A, and injects Plan A before prompts/tool calls. The nested project's explicitly active Plan B is ignored.

The injection repeats, so a correction from the user competes with the wrong plan again on the next hook event.

## Expected behavior

[...]

PR #216, opened 2026-08-07
refactor: merge language-variant skills into one canonical skill

### LINKED ISSUE #130: Consolidate language variants into a single skill with locale parameter
## Problem

The current plugin registers **6 separate skills** for language support:

- `planning-with-files` (English)
- `planning-with-files-ar` (Arabic)
- `planning-with-files-de` (German)
- `planning-with-files-es` (Spanish)
- `planning-with-files-zh` (Chinese Simplified)
- `planning-with-files-zht` (Chinese Traditional)

Each variant appears as a distinct entry in the user's skill list. For users who only need one language, the other 5 are dead weight — they clutter the skill list and consume system prompt tokens on every session (each skill's `description` is injected into context).

## Analysis

After inspecting v2.33.0:

| Aspect | Same across all 6? |
|--------|-------------------|
| Scripts (`session-catchup.py`, `init-session.sh`, etc.) | ✅ Byte-for-byte identical |
| Templates | ✅ Identical structure |
| Core logic in SKILL.md | ✅ Same rules, same workflow |
| SKILL.md prose & hook `command_message` | ❌ Translated per language |

The **only** difference is the natural-language text in SKILL.md and hook messages. The actual planning logic is identical.

## Why this is unnecessary

AI coding assistants (Claude Code, Cursor, Codex, etc.) already respond in the user's configured output language regardless of the skill's source language. A user with `output: zh-TW` configured will get Traditional Chinese responses from an English SKILL.md — the AI handles the translation naturally.

Duplicating the entire skill per language:
1. **Pollutes the skill list** — users see 5 skills they'll never invoke
2. **Wastes context tokens** — each skill description is loaded into the system prompt every session
3. **Increases maintenance burden** — any logic change must be replicated across 6 SKILL.md files

## Suggestion

Consolidate into a **single skill** (`planning-with-files`) with one of these approaches:

### Option A: Single SKILL.md, let the AI translate
Remove language variants entirely. The English SKILL.md already works for all languages because the AI translates its output. This is the simplest approach.

### Option B: Locale parameter
Keep one skill, add an optional `locale` argument:
```
/plan --locale zh-TW
```
If locale-specific hook messages are important, the skill can select the appropriate message strings at runtime.

### Option C: User-configurable locale in config
Add a `locale` field to the plugin config that controls hook message language, without duplicating the skill.

## Environment

- Plugin version: 2.33.0
- Platform: Windows 11 / Claude Code CLI

Thanks for the great plugin — the core planning workflow is solid. This is purely a packaging concern.

SomSamantray:
## What changed

Removed `skills/planning-with-files-{ar,de,es,zh,zht}/` and kept a single canonical skill. `/plan-ar`, `/plan-de`, `/plan-es`, `/plan-zh`, and `/plan-zht` are unaffected — each now invokes `planning-with-files:planning-with-files` directly with an inline instruction to write the planning files and respond in that language, instead of invoking a dedicated per-language skill. `scripts/bump-version.py` and the three tests that hardcoded the 5 removed paths (`test_skill_md_version_parity.py`, `test_skill_hook_dispatch_parity.py`, `test_stop_hook_dispatch.py`) were updated to match, the plugin/marketplace description now describes language support via the commands rather than separate skills, and README/AGENTS.md no longer reference the removed skill ids. Version bumped to 3.10.0 with a CHANGELOG entry documenting the removal as a breaking change.

## Why it changed

Issue #130 pointed out that the 6 skills only ever differed in prose — the assistant already responds in the user's language regardless of the skill's source language, so the 5 extra skills were pure duplication: 5 more descriptions injected into every session's system prompt, and one more place any logic change had to be replicated.

Checking that claim against the current `v3.9.0` tree found the cost had grown worse than described, not milder. The 5 variants had quietly drifted out of feature parity with the canonical skill:

- `skills/planning-with-files-ar/scripts/session-catchup.py` (and the other 4 variants) never got the `configure_utf8_stdio()` fix from v3.2.0 for the Windows encoding bug where non-ASCII session content crashed the hook silently — the same bug discussed in #19, still live for non-English users on Windows.
- They're missing 12 script files the canonical skill ships (`attest-plan`, `gate-stop`, `ledger-append`, `ledger-summary.ps1`, `phase-status`, `plan-doctor`, `resolve-plan-dir.ps1`, `set-active-plan`) — the completion gate, ledger, and plan-doctor features never reached non-English users at all.
- `scripts/sync-ide-folders.py`, the tool that keeps the IDE mirrors (`.codex/`, `.cursor/`, etc.) in parity with the canonical skill, was never extended to cover the 5 language directories, so this drift was structural, not a one-off oversight — every future feature would keep missing them.

Consolidating removes the duplication that makes that drift possible, rather than trying to backport the missing scripts and hope the sync tooling gets extended to catch the next one.

A few decisions worth flagging for review:

- **No runtime `locale` parameter** (issue's Option B/C) — `SKILL.md` has no such mechanism in this plugin format, and every hook-emitted diagnostic string is already English-only and byte-identical across the canonical skill and every "translated" variant today (verified by diffing `inject-plan.sh` between canonical and `-ar`: 0 lines different). There's no genuine locale-specific runtime text left to parameterize.
- **Kept the 5 commands instead of removing them** (issue's literal Option A) — this preserves every entrypoint the README documents today (`/plan-ar` etc.) for a one-line diff per command file, rather than a user-facing removal.
- **Deleted the 5 skill directories outright rather than a deprecation stub** — a stub with no hooks would still need threading through the hook-dispatch and version-parity tests for a release cycle, for an artifact with no functional value once the canonical skill covers every language. This repo's own convention for a breaking change is a CHANGELOG entry (e.g. `3.9.0`, `2.38.1`), not a redirect stub.

## How I tested it

`python -m pytest tests/ -q` passes clean: 411 passed, 11 skipped. `python scripts/bump-version.py 3.10.0 --dry-run` confirmed the trimmed 14-entry parity set resolves with no missing-file errors before I ran the real bump. Manually re-verified with `grep` that no live reference to a removed skill id remains outside `CHANGELOG.md`'s historical entries.

I don't have a way to invoke `/plan-ar` etc. end-to-end from this environment (that requires a live Claude Code session), so the one thing I can't verify directly is that each rewired command still produces output in its target language — I re-read all 5 command files and the instruction line is present and unchanged in intent from before, but a maintainer spot-check of at least one non-English command before merge would be good given this is the change's core user-facing guarantee.

## Known limitations

- `python scripts/bump-version.py` reports `MISSING: clawhub-upload/SKILL.md` on a fresh clone — pre-existing behavior (that directory is gitignored publish staging), unrelated to this change and not touched here.

Closes #130.


OthmanAdi:
@SomSamantray Not merging this one, and the reason is not the quality of the work. The audit in here is the most useful thing anyone has sent this repo in a while, and v3.10.0 acts on all of it.

Everything I could check from your diff held up. The five variants really were missing `configure_utf8_stdio()`, they really shipped 8 of the 20 scripts the canonical skill ships, and your list of the 12 missing ones matched exactly. `sync-ide-folders.py` really did cover only the three hook dispatch targets from #212, which is what made the drift structural rather than a one-off.

Two things decided against the deletion itself.

**The install numbers, and what deletion actually does to them.** The five variants carry 39.8K of the project's 89.3K installs on skills.sh. I checked what removing the directories would do there by reading the `vercel-labs/skills` source and issue tracker rather than assuming. Discovery is a live directory scan, so the CLI would immediately stop finding them, but the website has no pruning job: there are two open feature requests asking for one (#415 and #416, both unresolved for months) and a run of reports from repos in exactly this position, including #1553, a repo that deleted language-suffixed directories after a rename and could not get the entries taken down. So deleting the directories would not have removed those five listings. It would have left five permanent entries advertising an install command that errors, which is worse than either keeping or removing them.

**A behavior risk in the command rewrite.** `plan-de.md` tells the model that all planning files must be in German, and the canonical templates it now reads are English. Nothing in the canonical `SKILL.md` or templates tells the model to leave `**Status:** in_progress` alone, and `check-complete.sh` matches that token with `grep -F`. A model that reasonably translates that field along with the headings silently disables the completion gate for non-English users, which is the population the change is meant to help. The existing translated templates avoid this because a translator made that call once, correctly: the status tokens are literal English in every one of them, and the prose around them is not.

What shipped instead in v3.10.0, from your findings:

- The 12 missing scripts now reach all five variants. 60 files created, 0 overwritten.
- `configure_utf8_stdio()` backported into all five `session-catchup.py` by insertion, so the translated prose in those files is untouched.
- `sync-ide-folders.py` now records which variant scripts are translator-owned. Its old comment claimed variant scripts were language-neutral, which is false and would have caused real damage if a later maintainer had trusted it and flipped `include_scripts=True`: `-de/session-catchup.py` contains German prose and five `-ar` scripts contain Arabic.

Your two AGENTS.md corrections were right as well, dropping the `.kiro` row and pointing `.pi` at its `package.json`. I did not take them from this branch since they are tangled with the deletion commits, but send them as a small standalone PR and it goes in with your name on it.

You are in CONTRIBUTORS.md for the audit. #130 stays open with the install numbers attached, since the consolidation question is real even though this particular answer to it is not the one this repo can take.

OthmanAdi:
@SomSamantray Following up because this thread is now closed properly. #130 shipped in v3.11.0, and your audit is part of why it could.

The route that worked was moving the five variants one directory deeper rather than deleting them, so the plugin scan stops registering them while every install command keeps resolving. Your work is what made that safe to attempt: you proved the drift #130 predicted, named the twelve missing scripts and the missing Windows UTF-8 fix exactly, and showed the sync tooling covered only three dispatch targets. All of that shipped as v3.10.0. Had it not, this release would have moved five directories that were quietly a subset install.

You are credited in the v3.11.0 release notes alongside everyone else who pushed on this.

https://github.com/OthmanAdi/planning-with-files/releases/tag/v3.11.0

Issue #191, opened 2026-07-01
Codex hooks report 0/0 phases for an unstructured task_plan.md


mixian939:
Hi, thank you for maintaining this project.

I am mainly a Codex user and not very familiar with this codebase. I noticed this because the hook warning kept appearing in my Codex conversations and consuming context/tokens. I asked Codex to help me reduce the issue and identify the likely cause, so please treat this as an end-user bug report with a suggested fix, not as a confident maintainer-level patch.

## Environment Where I Saw This

- OS: Windows
- Client: Codex desktop app
- Shell: PowerShell
- Skill/integration: `planning-with-files`
- Hook type involved: Codex hooks, especially the Stop hook
- Local workspace path style: Windows paths such as `C:\Users\...\Documents\app-code`
- Situation: a workspace had a `task_plan.md` file, but it was not a structured planning-with-files plan with `### Phase ...` headings

The exact message I saw from my local Codex hook adapter was:

```text
[planning-with-files] Task incomplete (0/0 phases done). Update progress.md, then read task_plan.md and continue working on the remaining phases.
```

In the current upstream shell/PowerShell scripts, the equivalent advisory wording appears to be closer to:

```text
[planning-with-files] Task in progress (0/0 phases complete). Update progress.md before stopping.
```

The wording differs, but the bug shape is the same: the hook reports planning progress even though there are zero phases and no real active structured plan.

## Trigger Conditions

This seems to happen when all of these are true:

1. The Codex planning hooks are installed/enabled.
2. The workspace contains a file named `task_plan.md`.
3. That file does not contain any structured planning phase headings such as:

   ```markdown
   ### Phase 1: Requirements & Discovery
   ```

4. A hook runs that treats the existence of `task_plan.md` as enough to mean "there is an active plan".

A minimal `task_plan.md` that can trigger the issue:

```markdown
# Task Plan
```

or any unrelated project file named `task_plan.md` with no `### Phase` headings.

## Actual Behavior

The hooks treat the unstructured file as an active planning session.

Then the phase counter finds:

```text
TOTAL=0
COMPLETE=0
IN_PROGRESS=0
PENDING=0
```

The Stop hook/advisory logic then emits a message containing:

```text
0/0 phases
```

From a user perspective this looks like Codex is repeatedly saying there is an unfinished planning task, even though no real planning-with-files phase exists.

It is especially annoying in Codex because hook output is inserted into the conversation/context, so the same false warning can repeatedly waste tokens.

[...]

PR #222, opened 2026-08-18
fix: ignore unstructured plans in Copilot PowerShell stop hook

Whxuan0701:
## What changed

- Added a zero-phase guard to the GitHub Copilot PowerShell AgentStop hook, matching the existing shell behavior.
- Added a behavior test that executes the real PowerShell script against an unstructured task_plan.md fixture.

## Why

A task_plan.md without any ### Phase headings produced a misleading "Task incomplete (0/0 phases done)" message on Windows. The shell hook received this guard in the issue #191 fix, but the PowerShell counterpart was omitted.

## How tested

- python -m pytest tests/test_copilot_agent_stop.py -q (1 passed)
- python -m pytest tests/ -q (430 passed, 6 skipped, 474 subtests passed)

## Known limitations

The PowerShell behavior test skips on hosts without a PowerShell executable. The repository Windows CI job provides Windows PowerShell and exercises it.

### DIFF
--- a/.github/hooks/scripts/agent-stop.ps1
+++ b/.github/hooks/scripts/agent-stop.ps1
@@ -32,6 +32,11 @@ if ($COMPLETE -eq 0 -and $IN_PROGRESS -eq 0 -and $PENDING -eq 0) {
     $PENDING = ([regex]::Matches($content, "\[pending\]")).Count
 }
 
+if ($TOTAL -eq 0) {
+    Write-Output '{}'
+    exit 0
+}
+
 if ($COMPLETE -eq $TOTAL -and $TOTAL -gt 0) {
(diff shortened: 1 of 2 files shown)


OthmanAdi:
@Whxuan0701 Merged as c361552 and shipped in v3.10.2.

The diagnosis holds. The #191 guard reached `agent-stop.sh` and the canonical scripts and stopped there, so the PowerShell twin kept answering `Task incomplete (0/0 phases done)` on a plan that was never phase structured.

Your diff prompted a sweep of every script in the repository that emits an `(N/M phases done)` message. One more copy had the same gap: `.cursor/hooks/stop.ps1` was auto continuing on `0/0`. Guarded in ee6021b, so no unguarded copy is left.

The test is what made this straightforward to accept. It runs the real hook under real PowerShell rather than asserting on file contents, and I confirmed it fails against the pre merge script and passes after.

You are in CONTRIBUTORS.md.

Issue #195, opened 2026-07-06
Stop-hook `decision: block` + cwd attachment still hijack unrelated one-shot `codex exec` sessions (post-#178, v2.41.0)


marcmuon:
## Summary

Two unrelated **one-shot `codex exec` workflows** that merely *share a cwd* with an incomplete planning-with-files plan get hijacked by the Codex hooks — even on **v2.41.0**, which already includes the #146 session-attachment fix. Root cause is the same as #178 (the Stop hook emits `decision: block` for incomplete plans) combined with **cwd/root-`task_plan.md` attachment** that still auto-attaches every session in the directory.

This is distinct from #178 (closed) and #146 (fixed): those addressed the interactive planning loop and cross-session context leak. This report is specifically about **non-planning, one-shot `exec` sessions** (a CI code-review bot, a read-only research/analysis agent, a sub-orchestrator) that never opted into the plan but get their output corrupted **and** silently mutate the plan files.

## Symptom A — one-shot code review

```bash
codex exec -o review.md '$code-review review this branch vs main …'
```
Run in a repo whose cwd contains `task_plan.md`/`progress.md`, the review agent "reconciles the plan" instead of reviewing:
- `review.md` (the `-o` capture) is a ~300-byte stub: *"Updated planning files: progress.md logged the review… task_plan.md added one completed review-only phase so the planning hook no longer reports 0/0. Verification: planning check now reports ALL PHASES COMPLETE (1/1)."*
- The actual review verdict + findings get written **into `progress.md`**, and a fabricated *"Phase — code review — complete"* section is **appended to `task_plan.md`**.
- Run log shows `SessionStart` / `UserPromptSubmit` hooks firing and the completeness check.

The same invocation in a cwd **without** planning files behaves correctly (full review → `-o`).

## Symptom B — one-shot read-only research

```bash
codex exec -C <repo> -s read-only '<market-research prompt>'      # also with a $-skill token
```
The hooks inject plan context and the session tries to advance the plan, then **stalls with `EPERM` writing `.omx/state`** under the read-only sandbox. Research output ≈ empty. Correct behavior is only recovered by running from a **neutral empty dir** with no plan present.

## Why v2.41.0 doesn't already prevent this

[...]

PR #223, opened 2026-08-18
fix: honor PLANNING_DISABLED in GitHub Copilot hooks

Whxuan0701:
## What changed

- Added PLANNING_DISABLED=1 guards to all GitHub Copilot shell and PowerShell hook entry points.
- Kept the PreToolUse permission decision as allow while suppressing planning context.
- Extended the existing opt-out regression suite to execute all ten Copilot hooks.

## Why

The documented per-invocation opt-out covered the canonical and Codex hook routes, but the GitHub Copilot integration still read task_plan.md and emitted planning reminders when PLANNING_DISABLED=1 was set. This could attach an unrelated plan to a one-shot Copilot task sharing the same working directory.

## How tested

- python -m pytest tests/test_planning_disabled_optout.py -q (14 passed, 10 subtests passed)
- python -m pytest tests/ -q (431 passed, 6 skipped, 484 subtests passed)
- sh -n .github/hooks/scripts/*.sh

## Known limitations

Only the exact value PLANNING_DISABLED=1 disables hooks, matching the existing repository contract. Other values preserve current behavior.

### DIFF
--- a/.github/hooks/scripts/agent-stop.ps1
+++ b/.github/hooks/scripts/agent-stop.ps1
@@ -8,6 +8,11 @@ $OutputEncoding = [System.Text.UTF8Encoding]::new($false)
 [Console]::OutputEncoding = [System.Text.UTF8Encoding]::new($false)
 $InputData = [Console]::In.ReadToEnd()
 
+if ($env:PLANNING_DISABLED -eq '1') {
+    Write-Output '{}'
+    exit 0
+}
+
 $PlanFile = "task_plan.md"
 
 if (-not (Test-Path $PlanFile)) {
--- a/.github/hooks/scripts/agent-stop.sh
+++ b/.github/hooks/scripts/agent-stop.sh
@@ -7,6 +7,8 @@
 # Read stdin (required — Copilot pipes JSON to stdin)
 INPUT=$(cat)
 
+[ "${PLANNING_DISABLED:-}" = "1" ] && { echo '{}'; exit 0; }
+
 PLAN_FILE="task_plan.md"
 
 if [ ! -f "$PLAN_FILE" ]; then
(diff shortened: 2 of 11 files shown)


OthmanAdi:
@Whxuan0701 Merged as 5898865 and shipped in v3.10.2.

The gap was real and wider than the PR scoped it. The #195 opt-out lives in `scripts/inject-plan.sh`, which is why the canonical and Codex routes picked it up in v3.9.0, and the Copilot hooks read `task_plan.md` directly, so nothing ever handed them the variable. Reading your diff sent me through every hook entry point in the repository. The eight Cursor hooks have the same shape and the same gap. They are guarded in the same release, ee6021b. `.gemini` is deliberately behind and stays uncovered.

Two changes on top of your commit.

`PreToolUse` emitted `permissionDecision: allow` on the disabled path. With no `task_plan.md` present the pre merge hook answered `{}` and yours answered `allow`, so setting `PLANNING_DISABLED=1` to make the skill inert also gave every Copilot tool call a blanket approval it would not otherwise have had. Both copies now emit `{}`, which is what the hook's own no plan file path already returns.

The tests only exercised the disabled case. I gutted all ten hooks to emit `{}` unconditionally and the suite stayed green. That is the same silent death shape as the v3.6.0 realpath failure and the v3.8.0 Stop hook, so the tests now run each hook with the variable unset and again with it set, and fail if the unset run is already inert. That baseline caught a separate bug on its first run: `error-occurred.ps1` read stdin into `$input`, which is PowerShell's automatic pipeline variable. Under `-File` the assignment does not stick, so the JSON parse always saw an empty string and that hook had never logged an error on Windows, under pwsh 7 or Windows PowerShell 5.1.

Suite 424 to 430. You are in CONTRIBUTORS.md.

PR #224, opened 2026-08-18
test: run Hermes probe with the active Python interpreter

Whxuan0701:
## What changed

- Changed the Hermes injection determinism probe to launch with sys.executable.
- Kept the probe and all assertions unchanged.

## Why

The test suite can be started successfully with python3 on systems that do not provide a python command alias, including a default macOS setup. The Hermes probe then failed with FileNotFoundError because it hard-coded python instead of reusing the active test interpreter.

## How tested

- python3 -m pytest tests/test_injection_determinism.py::NoRawWallClockTests::test_hermes_adapter_normalizes_wall_clock -q (1 passed with no python alias on PATH)
- python3 -m pytest tests/ -q (408 passed, 27 skipped, 474 subtests passed)

## Known limitations

This is a test portability change only. It does not change the Hermes adapter or runtime interpreter requirements.

### DIFF
--- a/tests/test_injection_determinism.py
+++ b/tests/test_injection_determinism.py
@@ -33,6 +33,7 @@ from __future__ import annotations
 import os
 import shutil
 import subprocess
+import sys
 import tempfile
 import unittest
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


OthmanAdi:
@Whxuan0701 Merged as 79ab59b and shipped in v3.10.2.

The probe spawned a hardcoded `python`, so on a host that ships only `python3` it raised `FileNotFoundError` instead of testing anything. `sys.executable` is the right handle.

Independent confirmation arrived the same week: PR #226 was opened from a macOS checkout and reported this as the single red test in an otherwise clean run, with the same root cause.

You are in CONTRIBUTORS.md.

PR #226, opened 2026-08-19
refactor: move language variants under skills/i18n so the plugin registers one skill

### LINKED ISSUE #130: Consolidate language variants into a single skill with locale parameter

dylanpulver:
## What changed

The five language variants moved from `skills/planning-with-files-<lang>/` to `skills/i18n/planning-with-files-<lang>/`. Nothing was deleted, merged or renamed. All five skill names, all five `npx skills add --skill` commands in the README, and all five install destinations are exactly what they were.

One directory of depth is the whole mechanism. Claude Code discovers a plugin's skills by scanning `skills/*/SKILL.md` at a single level and does not recurse, so a variant one level down is not registered on the plugin route. `npx skills add` resolves `--skill` by skill name across a recursive scan of the repository, so the skill route does not care about the depth at all.

The five language slash commands can no longer name a plugin-registered skill, so `/plan-de` and its four siblings now read the translated `SKILL.md` from `$HOME/.claude/skills/planning-with-files-de/SKILL.md` first, then from `$HOME/.claude/plugins/marketplaces/planning-with-files/skills/i18n/planning-with-files-de/SKILL.md`, which is the same marketplace path the hook dispatch blocks already probe. That keeps the translated body and the translated templates in play on the plugin route instead of falling back to canonical English. Each of the five also now states that the status tokens stay literal English because `check-complete.sh` matches them with `grep -F`. That is the failure you flagged on #216, written into the file rather than left to the model to infer.

Counts: 5 directories moved (120 files), 15 files edited, 1 test file added.

## Why it changed

Your August comment settled two of the three costs in #130 and left one open, the system prompt cost of five extra descriptions per session on the plugin route, to be attacked without deleting anything.

`claude plugin details planning-with-files` prices it. On v3.10.1 the plugin projects about 2,254 always-on tokens per session across 19 components, and the five variants are about 1,010 of those: roughly 300 for es, 280 for de, 210 for ar, 110 for zh, 110 for zht, against roughly 180 for the canonical skill itself. Every plugin user was paying more for the five languages they did not pick than for the skill they installed. After the move the same command reports 14 components and about 1,042 always-on tokens.

## The backward compatibility decision

I kept the variants installable at their current names rather than at a fixed depth, because your #216 review was right that deletion freezes those five skills.sh listings while the install command behind each one starts failing. Moving them keeps every listing's command resolvable, so nothing on that side goes stale.

I checked the install route rather than assuming it. `npx skills add <repo> --skill planning-with-files-de` still finds all seven skills in the repository and still installs to `.claude/skills/planning-with-files-de/`, so the variant hook blocks that probe `$HOME/.claude/skills/planning-with-files-<lang>/scripts/` keep resolving as before.

The residual risk worth naming is that I can verify the CLI and not the skills.sh website indexer. If that indexer keys on `skills/<name>/` instead of a recursive scan, the five listings could go stale on the site even though every install command keeps working. If that risk is not one you want to take, say so and I will rework this to leave the directories exactly where they are and take the cost out of the description text instead. That version saves less, but it touches no discovery surface.

No version bump and no CHANGELOG entry here, since the release protocol owns both and three other PRs are open.

## How I tested it

`python -m pytest tests/ -q` gives 398 passed, 39 skipped, 348 subtests passed. The single failure, `test_injection_determinism.py::NoRawWallClockTests::test_hermes_adapter_normalizes_wall_clock`, is pre-existing and environmental: it shells out to `python`, this machine only has `python3`, and it fails identically on a clean `main`.

`python scripts/bump-version.py 3.10.2 --dry-run` resolves the parity set at the new paths with 18 changed and 0 errors. `python scripts/sync-ide-folders.py --verify` reports every IDE folder in sync with the five variant manifests rebased on `skills/i18n/`. `claude plugin validate . --strict` passes.

The before and after token numbers come from `claude plugin details` with this repository installed as a local marketplace, which is also how I confirmed the component count drops from 19 to 14.

A grep for the old paths is clean outside the `CHANGELOG.md` and `CONTRIBUTORS.md` history entries, which describe past releases and should keep their original paths. `tests/test_plugin_skill_surface.py` is new and fails if a variant is ever placed directly under `skills/` again, so this cost cannot quietly come back the way the script drift did.

What I could not check from here: I cannot run `/plan-de` in a live session, so the disk read in the five rewritten commands is verified by construction, meaning both candidate paths exist on their respective routes, rather than by execution. A spot check of one non-English command before merge would be worth your time.

## Known limitations

The five language commands cost roughly 150 more on-invoke tokens each, paid only when someone runs the command, in exchange for the always-on saving.

`scripts/_v240_update_hook_bodies.py` is a one-shot v2.4.0 migration that nothing calls. I repointed its five paths so a repository grep stays clean. Happy to leave it frozen as history instead.

Happy to reshape any of this, including the directory name if `i18n` is not what you want it called.

Fixes #130.


OthmanAdi:
@dylanpulver Reviewed in depth, not merged yet. The directory layout question belongs to me and I want to answer it deliberately, since this is the third attempt at #130 and the second one I have had to weigh against install numbers rather than code quality. What follows is what verification found, so you are not waiting on a black box.

**The diff itself is clean.** 120 of 120 file changes are `R100` renames, one added test, no mode changes, no new dependencies, no install hooks, nothing new in any install path. Zero lines of translated content altered. `sync-ide-folders.py --verify` exits clean at the new paths, `bump-version.py --dry-run` resolves the parity set, and the suite is 427 passed with 11 skipped on your branch.

**Your central claim checks out, and I verified it rather than reading it.** The skills CLI at v1.5.22 walks `skills/` to depth 3 and matches `--skill` against the frontmatter name or the directory basename, so `planning-with-files-de` still resolves at depth 2. An install into an isolated fake home landed at `~/.claude/skills/planning-with-files-de/` with the full 18 script surface, byte identical destination to before the move. The published command on skills.sh is the `--skill` form, not a path form, so the advertised command keeps working. That is the asymmetry with #216: deletion left a permanent listing advertising a command that errors, and a move does not.

**The plugin loader claim also checks out.** In Claude Code 2.1.87 the scan is a single `readdir` over `skills/` with a `SKILL.md` read per entry and no recursive call, and `plugin.json` declares no `skills` key so auto discovery applies. Six plugin skills register today; after the move it is one.

**One thing you did not find, and it is the reason I am pausing.** `~/.agents/.skill-lock.json` records `skillPath: "skills/planning-with-files-de/SKILL.md"` for every existing variant install. On `npx skills update`, `buildUpdateInstallSource()` appends that recorded subpath unconditionally for this repo, and after the move the subpath does not exist. Confirmed by pointing the CLI at both paths in your tree: the old one returns `Local path does not exist`, the new one resolves. So every existing variant install loses `skills update` until it is reinstalled. It is a soft failure, the installed skill keeps working and a fresh `--skill` install always works, and `skills update` may already be broken for this repository for unrelated reasons (vercel-labs/skills #809). No edit to this PR removes it. It is inherent to moving.

**Two corrections either way.** Candidate 2 in the five commands points at `$HOME/.claude/plugins/marketplaces/...`, but the plugin runs from `~/.claude/plugins/cache/<marketplace>/<plugin>/<version>/`, so on the plugin route that path reads the marketplace clone at master tip rather than the pinned version, and it does not exist at all under `CLAUDE_CONFIG_DIR` or the zip cache route. `${CLAUDE_PLUGIN_ROOT}` is substituted by the loader and this repository already uses it in command prose at `commands/plan-doctor.md:11`. Separately, `README.md:285` still advertises the five variant skill ids as model invocable, which stops being true after the move.

**Credit where it is due on the part that mattered most to me.** You carried the literal English status token instruction into all five commands and put it outside the branch so it applies on the fallback path too. That specific hazard is what sank #216, and closing it in the file rather than trusting the model to infer it is the right call.

The alternative I am weighing is a three line change instead of a 121 file one: declaring `"skills": ["./skills/planning-with-files"]` in `plugin.json`. The loader only auto scans when that key is absent, so declaring it registers exactly one skill while every directory, every lock entry and every listing stays where it is. It rests on a documented manifest field rather than on an undocumented scan depth, and it does not touch `skills update`. Your version buys the same reduction and additionally makes the layout self documenting, which has its own value.

I will come back with a decision rather than leaving this open indefinitely. Nothing here is a criticism of the work.

dylanpulver:
Both corrections are in, pushed as d7c499c.

The five commands now use `${CLAUDE_PLUGIN_ROOT}/skills/i18n/planning-with-files-<lang>/SKILL.md`, matching the substitution already used at `commands/plan-doctor.md:11`, so the cache route and `CLAUDE_CONFIG_DIR` route both resolve. `README.md:285` is rewritten: only `planning-with-files:planning-with-files` stays model invocable on the plugin route, and the translations are reached through `/plan-<lang>` or a standalone `npx skills add --skill` install.

On the lock file: you are right and I missed it. `skillPath` is recorded at install time, `buildUpdateInstallSource()` appends it unconditionally, and after a move that subpath is gone, so existing variant installs lose `skills update` until reinstalled. Nothing in this PR can prevent that. It is inherent to moving, and it belongs in your decision rather than in a patch.

On the `plugin.json` alternative, I went and measured it, and the answer turned out to be version dependent in a way that matters to your decision.

At 2.1.87 you are correct. The loader computes `J = !manifest.skills && exists(skills/)` and only assigns `skillsPath` when `J` is true, so declaring the key disables the default scan. I ran your exact proposal against your master tree on 2.1.87 and got 1 skill, loader line `skillsPath=none, skillsPaths=1`. Same result on 2.1.104.

On current Claude Code it no longer restricts. Same tree, same manifest, 2.1.236 reports `skillsPath=exists, skillsPaths=1` and loads 6 skills, so the declaration adds a path rather than replacing the scan. A synthetic four skill fixture behaves the same way, and a declared path outside `skills/` gives 5 rather than 1. The change landed somewhere between 2.1.104 and 2.1.223. I could not bisect it further because npm ships a native installer stub after roughly 2.1.11x, so there is no readable bundle in that range. The zod description still reads "in addition to those in the skills/ directory", which is stale against the 2.1.87 code and accurate against today's.

One instrument warning, since it is what makes this easy to get wrong. `claude plugin details` is not a valid way to check this. It performs its own `skills/` scan, ignores the manifest gate, and reads from an install time cache. My control case declared `"skills": ["./does-not-exist"]` and it still reported 4 skills with no warning. The loader's own debug output via `--debug-file` is the only reading I would trust.

For completeness I checked this PR's claim the same way on current Claude Code: loading the branch with `--plugin-dir` registers exactly 1 skill.

So the three line version does what you want on 2.1.87 and does nothing on a current install, while the move holds on both. That does not settle the `skills update` cost, which is real and still yours to weigh. I am happy either way and glad the status token instruction was the part that carried over.

OthmanAdi:
@dylanpulver Taking this one. Verified it against the real loader rather than the description: 6 registered skills down to 1, 19 components to 14, always-on cost roughly 2,254 to 1,042 per session, all thirteen commands intact. The skills CLI still finds all seven by name and a targeted `--skill` install lands in the same place with the full surface.

It needs a rebase on master. v3.10.2 shipped since you branched and bumped version strings inside the files you rename, so GitHub cannot resolve it. Sorry, that conflict is mine, not yours.

Two changes while you are in there:

1. In all five `commands/plan-*.md`, the second candidate path should be `${CLAUDE_PLUGIN_ROOT}/skills/i18n/planning-with-files-<lang>/SKILL.md`. The marketplace clone tracks the default branch while an installed plugin is version pinned, and it does not exist under `CLAUDE_CONFIG_DIR` or the zip-cache route. The loader substitutes that variable, and the repo already uses it in `commands/plan-doctor.md`.
2. `README.md:285` still lists the five variant skill ids as model invocable.

Ping me when it is pushed and it goes in as v3.11.0.

dylanpulver:
Rebased and pushed as fb24ddd. GitHub reports it mergeable again.

Your v3.10.2 bumps won. All five variant `SKILL.md` files read `version: "3.10.2"` at the new paths, and git's rename detection carried the change across on its own with no manual resolution needed.

The rename shape held exactly: 120 R100 renames out of 135 files, the same as before. I also diffed the PR's own patch before and after the rebase with index lines stripped, and it comes out byte identical, so nothing about what the change does moved.

Both corrections you asked for were already in the previous push. The five commands use `${CLAUDE_PLUGIN_ROOT}/skills/i18n/planning-with-files-<lang>/SKILL.md`, no `plugins/marketplaces` path remains anywhere under `commands/`, and `README.md` no longer advertises the five variant skill ids as model invocable.

The suite here is 400 passed, 43 skipped, 1 failed. The failure is `test_copilot_shell_hooks_skip_context_when_disabled`, and it fails identically on a clean master worktree, so it did not come from this branch. Since I had it open, the cause looks like this: `run_sh` invokes the hooks with `sh`, but `.github/hooks/scripts/error-occurred.sh` declares `#!/bin/bash`. Line 36 escapes the context through `json.dumps`, producing a literal `\n`, and line 37 emits it with `echo`. On macOS `/bin/sh` is bash in POSIX mode with `xpg_echo` on, so `echo` turns that back into a raw newline inside the JSON string and `json.loads` rejects it as an invalid control character. `/bin/sh -c 'echo "{\"x\":\"a\nb\"}"'` shows it directly. Not this PR's to fix, but worth knowing about.

OthmanAdi:
Merged and shipped as v3.11.0.

The rebase was clean and your read of it was right: git carried the v3.10.2 bumps across the renames on its own, the patch is byte identical before and after, and both corrections were already in.

Measured on the merged tree before tagging: `Loaded 1 skills from plugin`, down from 6. 19 components to 14, always-on cost roughly 2,254 to 1,042 per session, all thirteen commands intact. The skills CLI still finds all seven by name, and a `--skill planning-with-files-de` install lands in the same place with the full 20-script surface and pulls in nothing else.

Your macOS diagnosis was correct and is fixed in the same release. Seven shell hooks built JSON by interpolating a `json.dumps` result and emitting it with `echo`; all seven now use `printf '%s\n'`. The repro you gave was the whole investigation, so that fix credits you in the changelog.

One thing this release adds that your PR surfaced indirectly: `docs/languages.md`. The translations had no documentation page at all, which mattered less while the plugin auto-registered them and is now the only way anyone finds them.

You are in CONTRIBUTORS.md and named in the release notes.

https://github.com/OthmanAdi/planning-with-files/releases/tag/v3.11.0

PR #228, opened 2026-08-21
fix: feed the Copilot error hook over a pipe so dash can parse it

dylanpulver:
Master CI has failed on the ubuntu leg for five consecutive runs. `error-occurred.sh` uses `<<<` twice, dash has no here-string, and `tests/test_planning_disabled_optout.py` invokes hooks through `["sh", script]`, so the `#!/bin/bash` shebang never applies.

The four sibling hooks pipe with `echo`. I did not copy that form. Under dash, builtin `echo` expands backslash escapes, and `$INPUT` is raw JSON, so an escaped `\n` inside an error message becomes a control character that `json.load` rejects. Feeding the hook `{"error":{"message":"line1\nline2 boom"}}`, `printf '%s\n'` emits the ErrorOccurred context and `echo` emits `{}`. The echo form would trade a loud parse failure for a silent one, which is why 3c78c3c moved the output side off `echo`.

## How this was tested

Reproduced the ubuntu condition locally with a `sh -> /bin/dash` symlink ahead of `PATH`:

- before: `AssertionError: 0 != 2 : ... error-occurred.sh: 32: Syntax error: redirection unexpected` at test line 200, matching the CI log
- after: `401 passed, 43 skipped, 357 subtests passed` under dash, same totals under macOS `/bin/sh`
- reverting the change reproduces the failure, so the suite does exercise it
- `dash -n` passes on all five scripts in `.github/hooks/scripts/`; before this change the file was the only one of the five that failed

## Limitations

Copilot invokes the hook as `bash` per `.github/hooks/planning-with-files.json`, so no user was affected. This is a CI and portability fix.

The four `.gemini/hooks/*.sh` carry the same `<<<`. No test runs them through `sh`, and `.gemini/settings.json` invokes them by path so their shebang applies, which is why they are out of scope here. Worth a separate pass if you want the construct gone repo-wide.

### DIFF
--- a/.github/hooks/scripts/error-occurred.sh
+++ b/.github/hooks/scripts/error-occurred.sh
@@ -21,7 +21,7 @@ for _p in /usr/bin/python3 /usr/local/bin/python3 /opt/homebrew/bin/python3; do
     [ -x "$_p" ] && { PYTHON="$_p"; break; }
 done
 [ -z "$PYTHON" ] && PYTHON=$(command -v python3 2>/dev/null || command -v python 2>/dev/null)
-ERROR_MSG=$($PYTHON -c "
+ERROR_MSG=$(printf '%s\n' "$INPUT" | $PYTHON -c "
 import sys, json
 try:
     data = json.load(sys.stdin)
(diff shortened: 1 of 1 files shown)


OthmanAdi:
@dylanpulver Merged as 95c903c, shipped in v3.11.1.

I reproduced this against real dash rather than reading it off the CI log. Pre-fix the hook exits 2 at line 32 with `Syntax error: redirection unexpected`. Post-fix it exits 0 and emits the ErrorOccurred context with the newline escaped as two characters. I also parsed all five scripts in `.github/hooks/scripts/` under a POSIX shell: `error-occurred.sh` was the only one that failed before, and none fail now.

Copying the sibling `echo` form would have cleared the syntax error and gone green in CI. Under dash it turns the escaped newline inside an error message into a real control character and `json.load` rejects it, so a loud failure becomes a silent one. I checked that separately: `echo` on the raw payload gives `Invalid control character`, `printf '%s\n'` round trips. Declining the fix that looks right, and writing why into the pull request, is what made this a one pass review.

Your scope note on `.gemini` holds. Those four are invoked by path in `.gemini/settings.json`, so their shebang applies, and `.gemini` is on the intentionally lagging list. Moving it is a separate decision.

You are in CONTRIBUTORS.md.

PR #229, opened 2026-08-21
docs: name the skill directory in the skills-only install commands

dylanpulver:
`docs/installation.md:101` and `docs/windows.md:38` are the only two install commands in the repo that copy `skills/*` instead of naming the skill directory. My #226 moved the language variants down a level, so that glob now also copies `skills/i18n/`, so `~/.claude/skills/i18n/` lands with no SKILL.md in it and the five translations sit at `~/.claude/skills/i18n/planning-with-files-de/`, one level below the path `/plan-de` probes first (`commands/plan-de.md:7`). Ran both forms into a scratch directory to confirm.

I did not add a second line copying `skills/i18n/*`, which would restore what the command used to install. `docs/languages.md:26` says installing English pulls in no translation, and dropping all six into `~/.claude/skills/` puts the five extra descriptions back into every session, which is the cost #226 removed on the plugin route. Section 4 of installation.md and `README.md:250` name `skills/planning-with-files`.

The guard went into `tests/test_plugin_skill_surface.py`, the file that owns the depth invariant. It matches lines beginning with `cp` or `Copy-Item` only, so prose about `skills/*/SKILL.md` does not trip it. Reverting either doc line fails it with both paths named.

Limitation: `docs/windows.md` carries no language-variant section, so a Windows reader on the skills-only route is not pointed at `npx skills add --skill`. Left alone rather than widened.

444 collected, 414 passed and 31 skipped locally; `sync-ide-folders.py --verify` clean.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -98,7 +98,7 @@ If you only want the skill without the full plugin structure:
 
 ```bash
 git clone https://github.com/OthmanAdi/planning-with-files.git
-cp -r planning-with-files/skills/* ~/.claude/skills/
+cp -r planning-with-files/skills/planning-with-files ~/.claude/skills/
 ```
 
 ---
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -98,6 +98,7 @@ If you only want the skill without the full plugin structure:
 
 ```bash
 git clone https://github.com/OthmanAdi/planning-with-files.git
+mkdir -p ~/.claude/skills
 cp -r planning-with-files/skills/planning-with-files ~/.claude/skills/
 ```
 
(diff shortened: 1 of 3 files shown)


OthmanAdi:
@dylanpulver Merged and shipped as v3.11.2: the two skills-only commands were still copying `skills/*` after the translations moved under `skills/i18n/`; your commit `3c5ea2b` now copies only `skills/planning-with-files`, the tracked-Markdown guard prevents that whole-directory shape from returning, both instructions create `~/.claude/skills` first so a fresh install cannot flatten `SKILL.md`, and PR #229 is included in your CONTRIBUTORS.md entry.

https://github.com/OthmanAdi/planning-with-files/releases/tag/v3.11.2

PR #243, opened 2026-09-15
fix(plans): accept BOM-prefixed active pointers

kuei51307-hub:
## Summary

- strip a leading UTF-8 BOM before validating `.planning/.active_plan`
- preserve the existing slug and path-safety rejection rules
- add cross-distribution coverage for BOM-prefixed active-plan pointers
- regenerate and verify all synchronized script copies

## Why

Windows editors and some PowerShell workflows can prepend a UTF-8 BOM. The POSIX shell implementation treated those bytes as part of the slug, so an otherwise valid active-plan pointer was rejected and the active marker disappeared.

## Tests

- focused plan-listing, active-plan, distribution, and sync tests: 137 passed, 13 skipped
- `python scripts/sync-ide-folders.py --verify`
- `git diff --check`


OthmanAdi:
Confirmed: the shell helper misses a valid active plan when `.active_plan` starts with a UTF-8 BOM. I reproduced this with no newline, LF and CRLF endings. Windows PowerShell 5.1 and `pwsh` already handle those inputs.

Your patch fixes the shell behavior without rewriting the pointer. I have approved the full CI run and am checking malformed pointers and all distribution copies before including this in v3.18.1. I am also adding regression coverage for the no-argument display path.

Thanks for the fix, @kuei51307-hub.

OthmanAdi:
Released in [v3.18.1](https://github.com/OthmanAdi/planning-with-files/releases/tag/v3.18.1) at 7c11cb75702a0d82f1ee8b6b1b632272cd84bb85.

@kuei51307-hub, your commit was merged with authorship preserved, and you are in CONTRIBUTORS.md. The shell helper now strips one leading UTF-8 BOM before the existing slug validation, restoring current-plan display and the active marker without rewriting the pointer.

The fix is synchronized across all 18 canonical shell helper copies. Supplemental regression tests cover display and listing with no newline, LF and CRLF endings. All five CI jobs passed on the final release commit: Linux, Windows and macOS pytest, plus Pi and OpenCode.

Thanks for the contribution.

PR #244, opened 2026-09-15
fix(sync): report missing canonical sources

kuei51307-hub:
## Summary

- detect missing enabled canonical sources during `--verify`
- print `MISSING SOURCE` with the source path and return a failure status
- continue scanning remaining mappings so one missing source does not hide other drift
- count affected sync entries and print an actionable recovery message
- add subprocess-level regressions covering read-only verification and unchanged sync behavior

## Why

A missing canonical source bypassed content comparison and could leave an IDE target reported as up to date even though correctness could not be established.

## Tests

- `python -m pytest tests/test_sync_ide_folders.py -q` — 15 passed
- `python scripts/sync-ide-folders.py --verify` — passed
- `git diff --check HEAD^..HEAD` — passed


OthmanAdi:
@kuei51307-hub I reproduced the missing-source false success against the parent revision. Removing one shared canonical template returned success there; this change returns failure and reports all eight affected mappings while still reporting other missing or drifted targets. The verification snapshots remained unchanged, and normal sync and dry-run behavior matched the parent.

The 15 focused tests and all five PR checks pass. I will merge this with contributor authorship preserved, include it in v3.18.2, and comment again when the release ships.

OthmanAdi:
@kuei51307-hub Merged and released in v3.18.2: f12e714e3fa28c8602c9e48e775bc843fe332914.

Verification previously reported success when a canonical source was missing. It now fails, reports every affected mapping, and continues checking other missing or drifted targets without modifying files. Independent before/after checks confirmed that normal sync and dry-run behavior stay unchanged.

Your commit keeps its original authorship. You are in CONTRIBUTORS.md, and your existing entry now includes #244. All five CI jobs passed on the final release commit.

PR #245, opened 2026-09-15
fix: isolate Python in adapter shell hooks

ShaunLinTW:
## What

Harden the remaining Codex, Gemini, and GitHub Copilot shell adapters by running their Python interpreters in isolated mode (`-I`). Add regression coverage that plants project-local stdlib-shadowing modules and verifies the adapters do not import them.

## Why

Python started with `-c` or from these shell entry points can otherwise place the project working directory on the import path. A project-local module such as `json.py` or `hashlib.py` can then shadow the standard library while a planning-with-files hook is running.

This follows the isolation hardening already applied to the main hook paths and closes the remaining shell-adapter gap noted in the changelog.

## Testing

- `python -m pytest tests/test_adapter_python_isolation.py -q` — 5 passed
- `python -m pytest tests/test_codex_hooks.py -q` — 20 passed
- `python -m pytest tests/test_planning_disabled_optout.py -q` — 17 passed
- `python -m pytest tests/test_automatic_catchup_callers.py -q` — 2 passed
- `bash -n` on all changed shell hooks — passed
- `git diff --check` — passed

## Limitations / platform notes

A local Windows full-suite run was also attempted. After working around this checkout's Git `safe.directory` ownership issue and enabling UTF-8 mode, it reached 137 passed / 9 skipped before stopping on an existing Windows `commandWindows` PostToolUse test outside this patch's changed call chain. The targeted and adjacent adapter suites above are green.


OthmanAdi:
@ShaunLinTW I reproduced the import-shadowing issue: project-local json.py and hashlib.py can execute through these adapter calls, and isolated mode prevents that. The diff adds no dependencies or install scripts.

The Windows review also found an encoding regression. With Python 3.12 and Git Bash, -I ignores PYTHONUTF8=1. A valid pinned project path containing Chinese characters is then rejected during PIN_REAL output, and the Copilot error hook can return an empty object for a Unicode error message.

I am adding explicit UTF-8 mode alongside isolation and regression coverage for both cases before merging. Your original commit and authorship will be preserved, with the compatibility fix in a separate maintainer commit. I will comment again when v3.18.2 ships.

OthmanAdi:
@ShaunLinTW Merged and released in v3.18.2: f12e714e3fa28c8602c9e48e775bc843fe332914.

The affected Codex, Gemini, and Copilot shell calls now use isolated Python, preventing project-local modules and Python environment overrides from shadowing standard-library imports. Your original commit is preserved. The separate maintainer correction adds explicit UTF-8 mode so Unicode project paths and error messages keep working on Windows, and the Gemini/Copilot regression tests use their declared Bash runtime.

You are in CONTRIBUTORS.md and the release Thanks section. The import-shadowing and Unicode regressions were reproduced before the fixes and verified afterward. All five CI jobs passed on the final release commit.

PR #247, opened 2026-09-16
fix: add PowerShell slug-mode plan initialization

ShaunLinTW:
## What changed

- add named-plan slug mode to `init-session.ps1`, matching the shell initializer's `.planning/<date>-<slug>/` flow
- add `-PlanDir` support with generated slugs and collision suffixes
- preserve zero-argument legacy root-mode behavior and output
- keep `-Template`, `-Autonomous`, and `-Gated` working with isolated plans, including root-mode policy-floor inheritance
- sync the canonical PowerShell initializer to the repository's shared English install surfaces
- add Windows PowerShell regression coverage and update affected docs/evaluation notes

## Why

The shell initializer already supported isolated named plans for parallel sessions, while the PowerShell initializer still wrote only to the project root. This closes that parity gap without changing legacy zero-argument behavior.

## Testing

- `tests/test_init_session_powershell_slug.py`
- `tests/test_init_session_slug.py`
- `tests/test_init_modes.py`
- `tests/test_canonical_script_sync.py`
- `tests/test_ps1_windows_encoding.py`
- sync verification via `scripts/sync-ide-folders.py --verify`
- `git diff --check`

Fresh focused verification: **29 passed**.

A fail-fast full-suite run on this Windows checkout reached **136 passed, 12 skipped, 1 failed** at `tests/test_codex_plugin_operations.py::CodexPluginOperationsTests::test_command_windows_runs_from_spaced_cached_root`. That existing Windows Codex hook test returns empty stdout in this environment. The failing test path and `.codex/hooks` execution path are unchanged by this PR.


OthmanAdi:
Confirmed against master 746f57a. Your `tests/test_init_session_powershell_slug.py` fails 8 of 9 cases with the baseline `init-session.ps1` (only the zero-argument legacy case passes, as expected) and all 30 focused tests pass on your branch, including `test_init_session_slug.py`, `test_init_modes.py`, `test_canonical_script_sync.py` and `test_ps1_windows_encoding.py`. The ten PowerShell copies are byte-identical.

One thing I will harden in a maintainer follow-up commit on top of your work: slug mode writes `.planning\.active_plan` with `Set-Content`, which follows a reparse point and truncates a hardlinked pointer in place. `scripts/set-active-plan.ps1` already has the contained atomic replacement for that file, so the initializer will route the pointer write through it and validate the `.planning` root before creating the plan, the same change #249 makes for `init-session.sh`. I will add PowerShell regressions for the hardlink, symlink and junction cases.

Full suite and an independent review are running now. This lands in the next minor release together with #248 and #249, and I will comment again when it ships, @ShaunLinTW.

OthmanAdi:

@ShaunLinTW, both of your commits were merged with authorship preserved, and you are in CONTRIBUTORS.md. `init-session.ps1` now has the same named-plan flow as the shell initializer: a positional name or `-PlanDir` creates `.planning\<date>-<slug>\`, collisions get a numeric suffix, the root `.mode` policy floor is inherited, and attestation is bound to the plan that was just created.

Maintainer follow-ups on top of your work, all in the release:

- The pointer write goes through `set-active-plan.ps1` and the planning root is verified with its listing check before anything is created (the PowerShell twin of #249). Your version wrote the pointer with `Set-Content`, which truncated a hardlinked `.active_plan` in place and created the plan through a junctioned `.planning` outside the project; both cases now have regressions.
- Slugs are built with `-creplace` and guarded to ASCII, because the case-insensitive `-replace` let a dotted capital I survive into a plan id that the resolvers then rejected.
- Plan files are written with `Out-File -LiteralPath`; with absolute paths, `-FilePath` failed under pwsh in a project path containing `[` or `]` while still printing "Created".
- Root `.mode` inheritance is case-sensitive like `inherit_root_mode` in the shell twin, an empty positional name stays in root mode, and `-PlanDir` without a name reports "untitled".

Known limitation noted in the CHANGELOG: the Cursor native PowerShell hooks still read only the root `task_plan.md`, so a named plan is not injected on that route yet.

PR #248, opened 2026-09-17
fix(catchup): require exact planning filenames

kuei51307-hub:
## Summary

- require `task_plan.md`, `findings.md`, and `progress.md` to match as exact normalized basenames
- reject lookalikes such as `draft_task_plan.md`, `findings.md.bak`, and paths with child components
- bound OpenCode SQLite candidates to exact planning paths and revalidate the decoded record
- apply the same boundary to canonical Claude/Codex/OpenCode recovery and the Hermes, MastraCode, and OpenCode adapters
- keep the canonical distributed catchup copies synchronized

## Why

Suffix matching allowed unrelated files such as `draft_task_plan.md` to become a session-recovery anchor. A later lookalike write could move the catchup window forward and omit context that occurred after the real planning update.

## Tests

- focused catchup and custom-adapter suites: 47 passed on Windows
- the same 47 focused tests passed in a Linux container
- IDE-folder synchronization verification passed
- all changed Python files compiled successfully
- independent clean-base reproduction, copy parity, diff, authorship, overlap, and secret-pattern checks passed


OthmanAdi:
Confirmed against master 746f57a. Running your new tests against the baseline scripts fails exactly where you describe: `planning_file_from_path("draft_task_plan.md")` returns `task_plan.md`, the Codex `patch_apply_end` lookalike counts as a planning update, and the Hermes, MastraCode and OpenCode SQLite scanners all anchor on lookalike names. On your branch the same 50 tests pass, the ten canonical copies are byte-identical and `sync-ide-folders.py --verify` is clean.

One gap I will close in a maintainer follow-up commit rather than ask you to redo: the repository has three more copies of this script that the sync tool never touches, the root `scripts/session-catchup.py` (the Claude Code plugin install path, still suffix-matching at line 344 and still using the `LIKE '%task_plan.md'` OpenCode query) and the five `skills/i18n/planning-with-files-*/scripts/session-catchup.py` translations. They will get the same exact-basename rule so every shipped copy behaves the same way.

Full suite and an independent review are running now. This lands in the next release together with #247 and #249, and I will comment again when it ships, @kuei51307-hub.

kuei51307-hub:
Thanks for the thorough verification and for catching the extra shipped copies outside the sync path. Appreciate you handling those in the maintainer follow-up.

OthmanAdi:

@kuei51307-hub, your commit was merged with authorship preserved, and you are in CONTRIBUTORS.md. Catchup now anchors only on an exact normalized basename, so a later write to `draft_task_plan.md` or `archive-progress.md` no longer moves the recovery window forward in the Claude, Codex, Hermes, MastraCode and OpenCode scanners.

Maintainer follow-ups in the same release:

- The root `scripts/session-catchup.py` (Claude Code plugin install path) and the five `skills/i18n` translations received the same rule, and `tests/test_session_catchup_copies_exact_basename.py` now loads every shipped copy so the boundary cannot drift again in one of them.
- The OpenCode query iterates the cursor lazily instead of `fetchall()`; write parts carry whole file bodies, and a session with many planning writes materialized all of them before the first validated row ended the loop.
- `test_lookalike_planning_filename_does_not_anchor_catchup` seeds a later part so it fails on the suffix-matching baseline; without it, the empty-parts early return let it pass either way.
- The CHANGELOG notes that OpenCode path matching is now case-sensitive, like the JSONL scanners already were.

PR #249, opened 2026-09-17
fix(init): replace active pointers safely

kuei51307-hub:
## Summary

- validate the physical `.planning` root before creating an isolated plan beneath it
- replace direct `.active_plan` redirection with the contained atomic update in `set-active-plan.sh`
- preserve a hardlinked peer instead of truncating its shared inode
- reject unsafe active-pointer links and planning roots that resolve outside the project
- synchronize the fix across all shipped slug-aware shell copies

## Why

Slug initialization wrote `.active_plan` with shell redirection. If that path was a hardlink, the write truncated the other linked file; if the planning root escaped the project through a link, initialization could create plan files outside the project. Reusing the selector's containment checks and atomic replacement removes both unsafe write paths.

## Tests

- init-session slug regressions: 9 passed and 1 skipped on Windows because file symlink creation is unavailable on this host
- all 10 init-session slug regressions passed in a Linux container
- all changed shell copies are byte-identical and pass `sh -n`
- IDE-folder synchronization verification passed
- independent clean-base reproduction, diff, authorship, overlap, and secret-pattern checks passed


OthmanAdi:
Confirmed against master 746f57a. I ran your three new tests with the baseline `scripts/init-session.sh`: the hardlink case overwrote the linked sentinel (`b'KEEP-ME' != b'2026-09-17-hardlink-test\n'`) and the junctioned `.planning` case exited 0 and created the plan outside the project. With your patch, 9 pass and the file-symlink case skips on this host because symlink creation needs a privilege here. All ten slug-aware shell copies are byte-identical and the eight copies without slug mode are correctly left alone.

Reusing `set-active-plan.sh` for both the root containment check and the atomic pointer replacement is the right call, since that script already carries the mktemp and mv path and refuses linked pointers.

I am running the full suite and an independent review now and will include this in the next release together with #247 and #248. I will comment again when it ships, @kuei51307-hub.

kuei51307-hub:
Thanks for reproducing the baseline failures and confirming the fix across the slug-aware copies. Appreciate the verification and follow-up.

OthmanAdi:

@kuei51307-hub, your commit was merged with authorship preserved, and you are in CONTRIBUTORS.md. Slug-mode initialization now validates the physical planning root through `set-active-plan.sh --list` before creating a plan and replaces `.active_plan` through the selector's contained atomic update, so a hardlinked pointer keeps its peer intact, a symlinked pointer is refused, and a planning root that resolves outside the project stops initialization. The fix ships across all ten slug-aware shell copies.

The PowerShell initializer received the same treatment in this release through `set-active-plan.ps1`, with matching hardlink, symlink and junction regressions.

PR #251, opened 2026-09-17
fix(cursor): resolve named plans in PowerShell hooks

### LINKED ISSUE #250: Follow-ups from the v3.19.0 release review (named-plan initializers, catchup adapters)
Edge cases found by the v3.19.0 release review that were not shipped with the release. Each item has a reproduction from the review; none affects the default zero-argument root mode.

1. PowerShell selector versus a read-only pointer. `set-active-plan.ps1 -VerifyRoot` accepts a `.active_plan` that carries the Windows ReadOnly attribute, but `[IO.File]::Replace` then fails with "Access to the path is denied", so `init-session.ps1` creates the plan directory and exits 1, leaving it behind. Either reject ReadOnly in `Test-SafeActiveFile` or clear the attribute before the replacement. The shell twin replaces a read-only pointer without error.

2. Concurrent PowerShell named-plan inits. Four simultaneous `init-session.ps1 "Race Case"` runs on Windows leave one to three of them failing with "could not safely update the active plan pointer" and can leave a `.active_plan~RF*.TMP` ReplaceFile artifact. The shell twin's `mv -f` succeeds. Retry `Replace`/`Move` on `IOException` and remove the artifact in the `finally` block.

3. `init-session.ps1` prints "Created task_plan.md" and "Planning files initialized!" after a denied write because `Out-File` errors are non-terminating (pre-existing, reproduced with an icacls deny under pwsh 7.6.6 and 5.1). Either `$ErrorActionPreference = 'Stop'` at the top or test `$?` after each write. The shell twin aborts under `set -e`.

4. Windows PowerShell 5.1 relocates its cwd to `$PSHOME` when the project path contains `[` or `]` (pre-existing). `set-active-plan.ps1` carries a recovery block for this; `init-session.ps1` does not, so root mode reports success without writing and named mode fails at `New-Item .planning`. Port the recovery block or refuse to run under 5.1 in such a cwd.

5. `init-session.sh` slugify keeps an embedded newline in a plan name, so `mkdir -p` creates a directory with a newline in its name, the selector then rejects the id, and init exits 1 with the directory left behind. Strip `\r\n` in `slugify` and validate `PLAN_ID` against the selector rule before `mkdir`. The PowerShell twin already replaces newlines.

6. `_format_opencode_part` in the root and `.opencode` catchup copies calls `state.get('input')` without an `isinstance` guard, and a non-JSON `data` row makes SQLite's `json_extract` raise in every copy (pre-existing, only reachable with corrupted OpenCode rows).

7. The Hermes plugin (`.hermes/plugins/planning-with-files/planning_files.py`) and the OpenCode plugin (`.opencode/packages/opencode-planning-with-files/src/core.ts`) still write `.active_plan` by truncation, the bug class #249 removed from the shell initializer.

8. Both initializers bind `PLAN_ID` during attestation but not `PWF_PLAN_ROOT`, so an inherited pin can redirect or block attestation of the plan that was just created. Set `PWF_PLAN_ROOT` to the project for the duration of the attest call.

9. The Cursor native PowerShell hooks (`.cursor/hooks/*.ps1`) read only the root `task_plan.md`, so a plan created by `init-session.ps1 <name>` is not injected on that route. Teach them to call `resolve-plan-dir.ps1`.

Sources: the three independent reviews and the adversarial verification run for v3.19.0 (PRs #247, #248, #249).

kuei51307-hub:
## Summary

Addresses item 9 from #250.

Cursor's native PowerShell hooks were still hard-wired to the root `task_plan.md`, so a named plan created with `init-session.ps1 <name>` could be active under `.planning/<slug>/` while the native Windows hook route injected, reminded, and stopped as if no selected plan existed.

## What changed

- Add one shared Cursor PowerShell plan-context helper and use it from `userPromptSubmit`, `preToolUse`, `postToolUse`, and `stop`.
- Resolve `PLAN_ID`, `PWF_PLAN_ROOT`, `.active_plan`, and the single-plan fallback through the existing `resolve-plan-dir.ps1` contract.
- Fail closed on invalid or ambiguous selection instead of falling back to a different root plan.
- Preserve the existing `PLANNING_DISABLED=1` behavior and the PreToolUse `{"decision": "allow"}` protocol response.
- Make the shared PowerShell resolver's `PWF_PLAN_ROOT` absolute-path check compatible with Windows PowerShell 5.1, where `Path.IsPathFullyQualified` is unavailable, then sync that canonical change to the shipped copies.
- Update Cursor docs and add native PowerShell regression coverage.

## Reproduction

Before this change, with:

```text
.planning/plan-a/task_plan.md
.planning/.active_plan -> plan-a
```

all four native Cursor PowerShell hooks missed the named plan. The user-prompt hook injected nothing, pre-tool did not reread it, post-tool emitted no plan reminder, and stop did not evaluate its phases.

## Validation

- focused native PowerShell/Cursor coverage: 17 passed
- PowerShell syntax parsing for the changed scripts: pass
- `python scripts/sync-ide-folders.py --verify`: pass
- all synced `resolve-plan-dir.ps1` copies are byte-identical
- `git diff --check`: pass

This intentionally does not close #250 because the issue tracks multiple independent follow-ups; this PR addresses item 9.


OthmanAdi:
@kuei51307-hub Verified on Windows PowerShell 5.1.26100. Against the v3.19.0 Cursor hooks your new test file fails 5 of 6 cases (only the legacy root case passes); against this branch all 6 pass.

The resolver side is confirmed as well: on v3.19.0, `resolve-plan-dir.ps1` under 5.1 with `PWF_PLAN_ROOT` set aborts with `Method invocation failed because [System.IO.Path] does not contain a method named 'IsPathFullyQualified'`, so the pin was dead on every 5.1 route that calls the shared resolver, not only in the Cursor hooks. All 15 shipped copies of the resolver on this branch are byte-identical to the canonical one.

The full suite is running on this branch and on the v3.19.0 baseline, and a review pass follows. I will comment again when it merges and ships.

OthmanAdi:
@kuei51307-hub Merged by rebase as 6031c38 and shipped in v3.20.0 (release commit 2fbbd77). You are in CONTRIBUTORS.md with this PR.

Two Opus review passes ran on it before the merge, with real probes under Windows PowerShell 5.1 and pwsh 7 against the canonical `inject-plan.sh` and `inject-plan.py`. Your change was correct on every "which plan gets injected" cell; the follow-ups I committed on top are in the same release:

- `[pscustomobject]` is rejected under ConstrainedLanguage mode, which turned every hook fire, the legacy root included, into "nothing injected" on WDAC machines. The shared context is a hashtable now.
- A stale `.active_plan` text, a dot-named directory or an invalid-slug directory beside a root plan made the route refuse where every canonical route serves the root plan. The scan applies the resolver's own filter and a stale pointer falls through; a pointer that is a directory or a link, a binding `PLAN_ID` that names nothing, and a plan the resolver refused for containment still fail closed (the last one silently, like `inject-plan.sh`).
- `resolve-plan-dir.ps1` used wildcard-interpreting `Test-Path` and `Get-ChildItem -Path`, so a project path containing `[` or `]` resolved nothing under pwsh; it is `-LiteralPath` throughout now. `attest-plan.ps1` carried the same `IsPathFullyQualified` call you removed from the resolver.
- A zero-byte pointer raised inside the in-process resolver call (`[string]` of an empty `Get-Content -Raw` is still `$null`); fixed in the four scripts that read the pointer.
- The hooks set the console output encoding to UTF-8; under the OEM code page the em-dash arrived as `-` and non-ASCII plan text as `?`.
- `pre-tool-use.ps1` prints its decision from a `finally` block, and the notices name the offending `PWF_PLAN_ROOT` or `PLAN_ID` value with the canonical wording.

The contract now runs under both interpreters in `tests/test_cursor_powershell_named_plans.py`, and `docs/cursor.md` describes the route as shipped. Thank you for the careful test file; it made the follow-ups easy to place.

kuei51307-hub:
Thanks for the thorough verification and follow-ups — glad the regression test helped. Appreciate the merge and release.

PR #265, opened 2026-09-17
fix: bind init attestation to current project

### LINKED ISSUE #261: Initializers bind PLAN_ID for the attestation call but not PWF_PLAN_ROOT, so an inherited pin redirects attestation
## Problem

Both initializers attest the plan they just created by calling the attester with `PLAN_ID` bound to the new plan (`init-session.sh` line `PLAN_ID="${PLAN_ID:-}" sh "${_attest}"`, `init-session.ps1` sets `$env:PLAN_ID = $PlanId` around the `attest-plan.ps1` call). Neither binds `PWF_PLAN_ROOT` for that call. The attester resolves through `resolve-plan-dir` and honours an inherited `PWF_PLAN_ROOT`, so a pin left in the environment from another project redirects or blocks the attestation of the plan that was just written in the current directory.

Scope: users who export `PWF_PLAN_ROOT` (the nested-project pin from issue #212) and then run the initializer in a different directory. With the variable unset there is no problem. The planning files themselves are created correctly in every case; only the attestation step is affected, and its failure is silent because both initializers swallow the attester's output.

## How to reproduce

```sh
export PWF_PLAN_ROOT=/path/to/other-project      # holds its own .planning
cd /path/to/this-project
sh scripts/init-session.sh "Fresh Plan"
sh scripts/attest-plan.sh --show                   # or inspect .planning/<id>/.attestation
```

Observed: the attester looks for `<date>-fresh-plan` under `/path/to/other-project/.planning`, prints `[plan-attest] PWF_PLAN_ROOT=... did not resolve to a project root holding a plan` (suppressed by the initializer), and the new plan carries no attestation. In root mode (no name argument) the inherited pin makes the attester attest the other project's root `task_plan.md` instead of the one just created.

## Suggested fix

Bind `PWF_PLAN_ROOT` to the current project root for the duration of the attest call, in the same place `PLAN_ID` is bound:

- `init-session.sh`: `PWF_PLAN_ROOT="$PWD" PLAN_ID="${PLAN_ID:-}" sh "${_attest}"` (the initializer already runs from the project root).
- `init-session.ps1`: save `$env:PWF_PLAN_ROOT`, set it to `(Get-Location).Path` (or the recovered project root) inside the existing `try`, restore it in the `finally` next to the `PLAN_ID` restore.

Keep the twins identical in behaviour. A regression test can export a pin to a second temporary project, run each initializer in the first, and assert the attestation file exists for the new plan.

## Origin

Found by the v3.19.0 release review (PRs #247, #248, #249). Tracked as item 8 of #250 before that issue was split into one issue per item.

TayfurYldz:
Closes #261

## What changed
- bind `PWF_PLAN_ROOT` to the initializer current project only while auto-attestation runs
- restore the inherited PowerShell environment value after attestation
- sync the canonical initializer changes to shipped adapters
- add shell and PowerShell regressions for an inherited stale plan-root pin

## Why
An inherited `PWF_PLAN_ROOT` could redirect the attester away from the project being initialized. The initializer already binds `PLAN_ID`; the project root now receives the same temporary binding.

## Testing
- `python -m pytest tests/test_init_modes.py tests/test_canonical_script_sync.py -q` -> 11 passed
- `git diff --check`
- full `python -m pytest tests/ -q` was also run on WSL: 868 passed, 116 skipped, 65 failed. The failures are Windows PowerShell/WSL path-interoperability tests (for example `set-active-plan.ps1` resolving `\wsl.localhost` paths), so I am not treating the full suite as green on this host.


OthmanAdi:
Verified: reproduced the claim on master with `PWF_PLAN_ROOT` pointing at a second temporary project (the plan is created, `.attestation` is missing), and the same run with this change rebased onto master writes the attestation.

One note on the branch as it stands. It is based on v3.19.0, whose `attest-plan.ps1` still called `[IO.Path]::IsPathFullyQualified`, a method Windows PowerShell 5.1 does not have. With the root pin now set during attestation that call runs on every autonomous init, so on 5.1 the attestation tests in `tests/test_init_session_powershell_slug.py` fail on the branch as is (4 failures here on Windows 11). v3.20.0 replaced that call with a drive regex; after the rebase onto master all 27 tests in `test_init_modes.py` and `test_init_session_powershell_slug.py` pass. The rebase merge takes care of it, nothing to change on your side, and the CI run I just approved may show the same Windows failures for that reason.

Merging once the full suite on the combined branch finishes. Release v3.20.1.

OthmanAdi:
One finding from the pre-release review, for the record. Binding `PWF_PLAN_ROOT` for the attest call in both modes leaves root-mode `--autonomous` and `--gated` plans unattested, on the shell and the PowerShell route alike: the attester only falls back to the legacy `./task_plan.md` when no selector is set and refuses the root plan under an explicit pin (`attest-plan.sh` line 60, `attest-plan.ps1` line 417), so init still printed "attested" while every hook then refused to inject the plan. No existing test covered root-mode attestation from init, which is why CI stayed green.

The PR is merged as is, with your commit as yours. One maintainer commit on top keeps your binding in slug mode, clears both selectors in root mode (which also stops an inherited `PLAN_ID` from redirecting the shell attester to a sibling plan, as the PowerShell twin already did), and adds root-mode regressions for both routes. Nothing to change on your side.

OthmanAdi:
@TayfurYldz merged with a rebase as 1bea197 and shipped in v3.20.1 (release commit faf1a15). Your binding stays as written for slug mode; the maintainer commit c05821a on top clears both selectors in root mode, for the reason in the comment above, and adds the root-mode regressions. Both routes now attest the plan that was just created whatever `PWF_PLAN_ROOT` or `PLAN_ID` the environment carries. You are in CONTRIBUTORS.md.

PR #266, opened 2026-09-17
fix: normalize multiline plan slugs

### LINKED ISSUE #257: init-session.sh: a plan name with an embedded newline creates an unusable plan directory
## Problem

`init-session.sh "<name>"` accepts a plan name that contains an embedded newline. `slugify` keeps the newline, `mkdir -p` creates a directory whose name contains it, the selector then rejects that id as an invalid slug, and init exits 1 with the directory left behind.

Scope: named-plan mode on the shell route (Linux, macOS, Git Bash) and only for a name that contains a newline, which normally comes from a script or an agent passing multi-line text as the argument. The default zero-argument root mode is not affected. The PowerShell twin already replaces newlines in `Get-PlanSlug`.

## How to reproduce

```sh
sh scripts/init-session.sh "$(printf 'line one\nline two')"
ls .planning/
```

Observed (Git Bash, 2026-09-17): a directory `.planning/<date>-line-one<newline>line-two/` is created with the three planning files inside it, then the run prints `Error: invalid plan ID. Use a named directory under .planning.` and `Error: could not safely update .../.planning/.active_plan.` and exits 1. The pointer is not updated and the malformed directory stays behind.

Cause: `slugify` in `skills/planning-with-files/scripts/init-session.sh` pipes the name through `tr '[:upper:]' '[:lower:]'` and `sed -e 's/[^a-z0-9]/-/g' ...`. `sed` works line by line, so the newline between the lines is never seen by the character class and survives into `SLUG`. `PLAN_ID` is then used in `mkdir -p "$PLAN_DIR"` before `set-active-plan.sh` validates it.

## Suggested fix

- Strip `\r` and `\n` in `slugify` before the `sed` step (`tr -d '\r\n'`), so the slug is a single line. This matches the PowerShell twin.
- Validate `PLAN_ID` against the selector's slug rule before `mkdir -p` (the same character set `set-active-plan.sh` accepts), and exit 1 with a clear message when it does not match, so nothing is created for an invalid name.

A regression test can pass a name with an embedded newline and assert exit code 0 with a single-line slug, or exit 1 with no directory created, depending on which behaviour is chosen. Making the slug single-line is the friendlier option.

## Origin

Found by the v3.19.0 release review (PRs #247, #248, #249). Tracked as item 5 of #250 before that issue was split into one issue per item.

TayfurYldz:
Closes #257

## Problem

The shell slugifier passed embedded CR/LF characters through to `sed`. Because `sed` processes input line by line, a newline survived slugification and became part of `PLAN_ID`. The initializer then created that malformed directory before the selector rejected the ID.

## Fix

Normalize CR/LF to `-` before the existing slug cleanup. This keeps multiline input on one line and matches the PowerShell slug behavior, where non-ASCII-alphanumeric characters become separators. The canonical script and generated IDE copies were synced together.

A regression test passes an embedded newline and verifies that initialization succeeds with one single-line plan directory and no newline in its name.

## Validation

- baseline reproduced on current `master`: multiline input exits 1 after creating a directory containing a newline
- `python -m pytest tests/test_init_session_slug.py tests/test_canonical_script_sync.py tests/test_bsd_userland_sim.py -q` -> 18 passed, 2 subtests passed
- `python scripts/sync-ide-folders.py --verify` -> all IDE folders in sync
- `git diff --check`
- full `python -m pytest tests/ -q` was also run: 868 passed, 116 skipped, 65 failed. The failures are the existing Windows PowerShell path-containment tests when Windows PowerShell is invoked against WSL `/tmp` paths (`planning directory ... outside the project or cannot be verified`); this change does not modify PowerShell code.


OthmanAdi:
Verified on master: `sh scripts/init-session.sh "$(printf 'line one\nline two')"` leaves `.planning/2026-09-18-line-one<newline>line-two/` behind and exits 1. With this change the same call exits 0 and creates `2026-09-18-line-one-line-two`. Ten copies updated, sync verify clean.

Merging with a rebase once the full suite on the combined branch finishes. Release v3.20.1.

OthmanAdi:
@TayfurYldz merged with a rebase as d03002b and shipped in v3.20.1 (release commit faf1a15). Root cause: `sed` works line by line, so the newline between the two lines never reached the character class and survived into `PLAN_ID` before `mkdir -p` ran. The `tr '\r\n' '--'` step is in all ten copies with the regression test. You are in CONTRIBUTORS.md.

PR #268, opened 2026-09-18
fix(hermes): require PLAN_ID for multiple plans

### LINKED ISSUE #264: Hermes plugin: several named plans without PLAN_ID resolve to the newest plan instead of requiring a selector
## Problem

Since v3.17.1 the shell, PowerShell and Python selection paths refuse to pick a plan when two or more named plans exist under `.planning/` and `PLAN_ID` is not set: the prompt-time hook prints `[planning-with-files] Multiple plans are available. Set PLAN_ID=<slug> for this session; nothing injected.` and every other point stays quiet (#240). A shared `.active_plan` pointer or a directory timestamp cannot tell which plan a session means, so guessing was removed.

The Hermes plugin (`.hermes/plugins/planning-with-files/planning_files.py`) still resolves the pointer and then the newest plan by modification time in that situation. Two Hermes sessions working on two named plans in the same project can therefore be injected with each other's plan after a pointer change, which is the shape #240 describes for Codex.

Scope: Hermes users with more than one named plan in one project and no `PLAN_ID` in the Hermes process environment. One named plan, a root plan, and any `PLAN_ID`-pinned session are unaffected.

## How to reproduce

```sh
sh scripts/init-session.sh "Plan A"
sh scripts/init-session.sh "Plan B"      # .active_plan now names plan B
# start a Hermes session in this project without PLAN_ID and send any prompt
```

Observed: the plan head of `Plan B` is injected. Expected (Claude Code, Codex, OpenCode and DeepSeek Harness routes): the one-line notice and no plan.

## Suggested fix

Port the rule from `inject-plan.sh` (lines 259 to 277 in v3.19.0): without `PLAN_ID`, count a root `task_plan.md` under an armed `.planning/sessions/` as one plan plus every valid slug directory that holds `task_plan.md`; when the count is above one, inject the notice from `pre_llm_call` once per turn and let `post_tool_call`, `pre_verify`, `/pwf-status` and the tools report that a selector is missing. The OpenCode and DSH plugins received this in v3.20.0 through their shared `core.ts` (`countSelectablePlans`), which is a compact reference for the Python port. `tests/test_hermes_first_class.py` already builds named plans and is the place for the regression.

## Origin

Found while porting the rule into the DeepSeek Harness plugin for #252.

jhuang-tt:
## Summary
- move Hermes multiple-plan selection into the shared resolver
- refuse active-plan / newest-by-mtime guessing when multiple named plans exist without PLAN_ID
- return the same selector-required message from prompt, status, and completion paths
- preserve single-plan, explicit PLAN_ID, session isolation, and PWF_PLAN_ROOT behavior

## Testing
- python -m unittest tests.test_hermes_first_class — 35 tests, 2 skipped
- python -m unittest tests.test_session_plan_ambiguity
- python -m unittest tests.test_unarmed_plan_ambiguity
- python -m py_compile on the changed Hermes modules and regression test

Closes #264


OthmanAdi:
@jhuang-tt  hey bud, why you closed this ?

PR #273, opened 2026-09-19
fix: tolerate malformed OpenCode replay parts

### LINKED ISSUE #269: session-catchup.py OpenCode replay still raises on null, array, non-string tool and object text part rows
## Problem

v3.20.1 (#258, PR #263) made the OpenCode adapter in `session-catchup.py` tolerate a part row whose `state` or `state.input` is not an object and filter non-JSON `data` in the planning-update query. The replay step after that query is still trusting: it reads every part row of the session with `SELECT data FROM part WHERE session_id = ? AND time_created > ?`, decodes it with `json.loads`, catches only `JSONDecodeError`, and passes the result to `_format_opencode_part`, which assumes an object with string fields. Four shapes still make the whole catchup run raise instead of skipping the row:

| `data` column | Error |
|---|---|
| `null` | `AttributeError: 'NoneType' object has no attribute 'get'` |
| `[]` (any JSON array) | `AttributeError: 'list' object has no attribute 'get'` |
| `{"type":"tool","tool":5}` | `AttributeError: 'int' object has no attribute 'lower'` |
| `{"type":"text","text":{"a":1}}` | `KeyError: slice(None, 300, None)` |

Scope: the explicit `session-catchup.py` command on projects that use OpenCode, only with corrupted or hand-edited rows. Normal OpenCode stores are well-formed. The Claude Code, Codex, Hermes and Pi catchup paths are separate code and not affected.

## How to reproduce

Point the adapter at a copy of an OpenCode store that already has a planning-file write, then:

```sql
UPDATE part SET data = 'null' WHERE id = '<a part id created after that write>';
```

and run `python scripts/session-catchup.py <project>`: `AttributeError: 'NoneType' object has no attribute 'get'`. The same with `'[]'`, `'{"type":"tool","tool":5}'` and `'{"type":"text","text":{"a":1}}'` gives the errors in the table. The function can also be called directly: `_format_opencode_part(json.loads('null'), 'ses_x')`.

## Suggested fix

- At the top of `_format_opencode_part`, return `None` when `data` is not a dict.
- Treat a non-string `tool` and a non-string `text` as empty (`tool = data.get('tool'); tool = tool.lower() if isinstance(tool, str) else ''`, same idea for `text`).
- Add the four rows to the fixture in `tests/test_session_catchup_opencode.py` next to the two from #258 and assert the run still reports the healthy `Tool edit` row.

Copies: the fix lands in `skills/planning-with-files/scripts/session-catchup.py` and is synced with `python scripts/sync-ide-folders.py`; the root `scripts/session-catchup.py` and `skills/i18n/planning-with-files-zh/scripts/session-catchup.py` are not covered by the sync tool and are updated by hand (`python scripts/sync-ide-folders.py --verify` and `tests/test_canonical_script_sync.py` check the result). Twelve copies carry the adapter in total.

## Origin

Found by the v3.20.1 release review of PR #263. Pre-existing behaviour, not introduced by that PR.

TayfurYldz:
## What changed
- ignore non-object OpenCode part rows instead of calling `.get()` on them
- treat non-string tool names and text payloads as malformed rather than raising
- cover null, array, numeric tool, and object text rows in the replay regression test

## Why
OpenCode stores part data as JSON without guaranteeing the decoded shape. `session-catchup.py` assumed an object with string fields, so malformed rows could abort replay instead of being skipped.

## Testing
- `python -m pytest tests/test_session_catchup_opencode.py tests/test_catchup_store_resolution_parity.py tests/test_session_catchup_copies_exact_basename.py tests/test_canonical_script_sync.py -q` -> 20 passed, 216 subtests passed
- `git diff 57fe376..HEAD --check` -> clean
- Full `python -m pytest tests/ -q` was also run. It reported 76 failures, all reproduced unchanged on the clean `57fe376` baseline; 896 passed and 147 skipped on the baseline.

Closes #269


OthmanAdi:
Confirmed against current master: null and array parts, a numeric tool name, and object-valued text can abort OpenCode replay. PR #273 skips those malformed fields while retaining healthy rows. Independent replay, privacy, and copy-parity checks passed. I plan to include this fix in the next patch release.

OthmanAdi:
@TayfurYldz, merged with your original authorship in 5ad545227689da67abd568acd0f8f2055f71e2d7 and released in [v3.20.5](https://github.com/OthmanAdi/planning-with-files/releases/tag/v3.20.5).

The replay formatter trusted decoded JSON shapes, so malformed rows could abort otherwise healthy sessions. Your guards handle those shapes without losing healthy replay rows. I reproduced all four failures on the previous release and verified the fix, including privacy and cross-project isolation checks.

Thanks for carrying the fix through the affected copies and adding the regression cases. You are in CONTRIBUTORS.md, with this contribution added to your existing entry.

The final release commit is 5ac39bcb41a9251d4253f7a2372b32147bfe6c2c. Its [CI run](https://github.com/OthmanAdi/planning-with-files/actions/runs/35642019298) passed the full pytest suite on Linux, Windows, and macOS, plus the Pi, OpenCode, and DSH plugin checks. ClawHub upload and npm publication are deferred.

